PR #11, opened 2022-12-08
Fix/Config number parsing

amit-momin:
- Using Number() to parse number configs
- Adjusted some TTL env var maxes to 1hr


alejoberardino (changes requested):
We should also add a check to the integer validator to make sure there's no decimals

amit-momin:
> We should also add a check to the integer validator to make sure there's no decimals

Looks like that validation already exists but it wasn't working with `parseInt`. But that was a good reminder to add tests for these scenarios. Included them in the last commit.

alejoberardino (approved):

PR #14, opened 2022-12-09
add validation for input overrides

karen-stepanyan:


src/validation/input-validator.ts
```diff
@@ -13,6 +16,7 @@ export class InputValidator {
 
   validateInput(input: Record<string, unknown>) {
     const normalized = this.initializeInputs(input)
+    this.validateOverrides(input?.['overrides'] as Record<string, Record<string, string>>)
```
alejoberardino, line 19: We should have types for these, something like
```typescript
type Override = Record<string, string>
type OverridesMap = Record<string, Override>
```

alejoberardino (changes requested):
Suuuper minor comment, LGTM otherwise

alejoberardino (approved):

PR #16, opened 2022-12-13
Rest & Batch consolidation

alejoberardino:
This one has very deep changes. I'll keep it as short as I can:
- `Rest` and `BatchWarming` transports have been consolidated into a new `Http` transport.
  - All requests will be processed in the background.
  - New interface allows for an arbitrary number of requests, to contemplate use cases like:
    - One request to the DP per each request to the EA.
    - A single batch request to the DP for all requests to the EA.
    - Any combination of requests, for example:
      - One for symbol batches, another for id batches.
      - Partial batches when the number of items to request would exceed the max for a batch endpoint.
- A new helper dependency `Requester` is added to centralize the execution of request to the Data Provider:
  - Makes rate limiting easy by processing requests sequentially and sleeping when capacity is reached.
  - Does not handle weights or priorities of any kind for now (good enough for most DPs).
  - Handles retries by adding requests back into the queue if they fail.
  - Coalesces requests by using a map alongside a linked list for the queue to ensure that repeated requests are not added as new ones.
- Background executor no longer relies on the rate limiter to sleep between calls. Now it only sleeps for a very small time (to allow for transports to implement their own timings)


src/background-executor.ts
```diff
@@ -69,13 +67,14 @@ export async function callBackgroundExecutes(adapter: Adapter, apiShutdownPromis
       } catch (error) {
         logger.error(error)
       }
-      const timeToWait = rateLimiter.msUntilNextExecution(endpoint.name)
+
+      // This background execute loop is no longer the one to determine the sleep between bg execute calls.
+      // That is now instead responsibility of each transport, to allow for custom ones to implement their own timings.
       logger.debug(
-        `Finished background execute for endpoint "${endpoint.name}", sleeping for ${timeToWait}ms`,
+        `Finished background execute for endpoint "${endpoint.name}", calling it again in 1ms...`,
```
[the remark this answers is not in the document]
boxhock: I see that SSE uses `trace` level for the exact same thing

boxhock (changes requested):
Looking really good, and looking forward to implementing EAs with these changes!

Just had a few small comments

boxhock (commented):
Finally spent some more time on the last (big) file

docs/porting-a-v2-ea-to-v3.md
```diff
@@ -81,36 +73,39 @@ export type EndpointTypes = {
...
-      method: 'GET',
-      params: {
-        from: req.base,
-        to: req.quote,
-      },
-    }
+    return params.map(req => ({
+      params: req,
+      request: {
+        baseURL: config.API_ENDPOINT || DEFAULT_API_ENDPOINT,
```
amit-momin, line 94: Probably something that we should have updated some time back but now that we've moved `API_ENDPOINT` to the EA side, don't we expect `config.API_ENDPOINT` to already have the default defined in it?
```suggestion
        baseURL: config.API_ENDPOINT,
```

docs/porting-a-v2-ea-to-v3.md
```diff
@@ -120,45 +115,63 @@ const restTransport = new RestTransport<EndpointTypes>({
...
-      baseURL: config.API_ENDPOINT || DEFAULT_API_ENDPOINT,
-      url: '/price',
-      method: 'GET',
-      params: {
-        from: params.map((req) => req.base).join(','),
-        to: params.map((req) => req.quote).join(','),
-      },
+      params,
+      request: {
+        baseURL: config.API_ENDPOINT || DEFAULT_API_ENDPOINT,
```
amit-momin, line 128: Same here
```suggestion
        baseURL: config.API_ENDPOINT,
```

docs/porting-a-v2-ea-to-v3.md
```diff
@@ -120,45 +115,63 @@ const restTransport = new RestTransport<EndpointTypes>({
...
-    //   {
-    //     params: { base: "BTC", quote: "USD" }
-    //     value: 123.45
-    //   }
-    // ]
+    return [
+      {
+        params: { base: "ETH", quote: "USD" }
+        response: {
+          result: data.price,
```
amit-momin, line 146: We didn't really do it better before but is it alright that this is a little ambiguous? Not a blocker to get this PR in but maybe we can make this example more realistic with request-response matching? Maybe we can just copy something in from coingecko.

src/transports/http.ts
```diff
@@ -0,0 +1,259 @@
...
+type ProviderRequestConfig<T extends HttpTransportGenerics> = {
+  /** The input paramters for requests that will get responses from the request in this struct */
+  params: T['Request']['Params'][]
+
+  /** The request that will be sent to the data provider to fetch values for the params in this struct */
+  request: AxiosRequestConfig<T['Provider']['RequestBody']>
+}
+
+/**
+ * Config object that is provided to the BatchWarmingTransport constructor.
```
amit-momin, line 50:
```suggestion
 * Config object that is provided to the HttpTransport constructor.
```

src/config/index.ts
```diff
@@ -334,6 +283,32 @@ export const BaseSettings = {
...
+    default: 1000,
+    validate: validator.integer({ min: 1, max: 10000 }),
+  },
+  BACKGROUND_EXECUTE_MS_HTTP: {
+    description:
+      "Time in milliseconds to sleep between HTTP transports' background execute calls, when there are no requests to send",
+    type: 'number',
+    default: 1000,
+    validate: validator.integer({ min: 1, max: 10000 }),
+  },
```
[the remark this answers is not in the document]
alejoberardino: I'll add a ticket to port over the notion page to the repo docs, so we can share it better and also do a review of the current state of the env vars

src/transports/http.ts
```diff
@@ -0,0 +1,259 @@
...
+   */
+  parseResponse: (
+    params: T['Request']['Params'][],
+    res: AxiosResponse<T['Provider']['ResponseBody']>,
+    config: AdapterConfig<T['CustomSettings']>,
+  ) => ProviderResult<T>[]
+}
+
+/**
+ * Transport implementation that takes incoming batches requests and keeps a warm cache of values.
```
amit-momin, line 92: Nitpick
```suggestion
 * Transport implementation that takes incoming batches of requests and keeps a warm cache of values.
```

src/transports/http.ts
```diff
@@ -0,0 +1,259 @@
...
+ * Within the setup function, adapter params are added to an set that also keeps track and expires values.
```
amit-momin, line 93: Nitpick
```suggestion
 * Within the setup function, adapter params are added to a set that also keeps track and expires values.
```

src/transports/http.ts
```diff
@@ -0,0 +1,259 @@
...
+ * In the background execute, the list of non-expired items in the set is fetched.
+ * Then, the list is passed through the `prepareRequest` function, that returns an AxiosRequestConfig.
+ * The Data Provider response is, they are passed through the `parseResponse` function to create a [[CacheEntry]] list.
```
amit-momin, line 96:
```suggestion
 * The Data Provider response is then passed through the `parseResponse` function to create a [[CacheEntry]] list.
```

src/util/requester.ts
```diff
@@ -0,0 +1,275 @@
...
+    await new Promise((unblock) => {
+      queuedRequest.promise = new Promise<RequesterResult<T>>((success, failure) => {
+        queuedRequest.resolve = success
+        queuedRequest.reject = failure
+        unblock(0)
+      })
+    })
+
+    // By the time we're here, we know that queuedRequest has both the unresolved promise, and the resolve and reject handlers within
+    // It's really, REALLY important for thread safety that from this point until this function returns, there are no async breaks
```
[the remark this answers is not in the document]
alejoberardino: I expanded on this one in the commit I'll upload shortly, lmk if it looks good

docs/porting-a-v2-ea-to-v3.md
alejoberardino: Could you expand on this?

amit-momin: Sure. It's just that the prepareRequest example is a realistic and looks like something that's used in an EA. But prepareResponse is sort of hard-coded response instead of an example. This isn't necessarily in the scope of the ticket though.

src/transports/http.ts
```diff
@@ -0,0 +1,259 @@
...
+    await this.responseCache.write(results)
+  }
+
+  private async makeRequest(
+    requestConfig: ProviderRequestConfig<T>,
+    adapterConfig: AdapterConfig<T['CustomSettings']>,
+  ): Promise<TimestampedProviderResult<T>[]> {
+    try {
+      const requesterResult = await this.requester.request<T['Provider']['ResponseBody']>(
+        requestConfig.params.map((p) => JSON.stringify(p)).join('|'),
```
amit-momin, line 200: Would there be any problems if this key gets extremely large especially for large batch requests?

src/util/requester.ts
```diff
@@ -0,0 +1,281 @@
...
+        logger.trace(`Request failed, sleeping for ${timeToSleep}ms...`)
+        await sleep(timeToSleep)
+
+        req.retries++
+        logger.trace(`Adding request to the queue (Key: ${key}, Retry #: ${req.retries})`)
+        this.queue.add(req)
+      }
+    } finally {
+      // Remove the request from our map
+      delete this.map[key]
```
amit-momin, line 276: If I'm understanding the logic correctly, wouldn't we want to only remove from the map if we've hit retries limit? We'd still want the request to be in the map if we'd added it to the back of the queue for a retry right?

test/index.test.ts
```diff
@@ -58,14 +58,13 @@ test('MTLS_ENABLED with no TLS params should error', async (t) => {
       'TLS_PRIVATE_KEY , TLS_PUBLIC_KEY and  TLS_CA environment variables are required when MTLS_ENABLED is set to true.',
     )
   }
-  // const api = await expose(adapter)
+  // Const api = await expose(adapter)
```
amit-momin, line 61: Not relevant to the ticket but why don't we just remove this commented out code completely?

docs/porting-a-v2-ea-to-v3.md
amit-momin: Maybe something like this? Unless that's less helpful to visualize the method.
```parseResponse: (params, res) => {
      return params.map((req) => ({
        params: req,
        response: {
          data: res.data,
          statusCode: 200,
          result: res.data.price,
        },
      }))
    },```

src/util/requester.ts
alejoberardino: correct, I'll make the change

src/util/requester.ts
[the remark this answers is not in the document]
boxhock: Yes, thank you

boxhock (approved):

PR #18, opened 2022-12-15
Feature/Add limit to subscription set

bogos:
[PDI-131](https://smartcontract-it.atlassian.net/browse/PDI-131)

[PDI-131]: https://smartcontract-it.atlassian.net/browse/PDI-131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/subscription-set/expiring-sorted-set.ts
```diff
@@ -16,28 +19,83 @@ interface ExpiringSortedSetEntry<T> {
...
+    this.list = new DoubleLinkedList()
+  }
 
   add(key: string, value: T, ttl: number) {
-    this.map.set(key, {
-      value,
-      expirationTimestamp: Date.now() + ttl,
-    })
+    if (this.map.has(key)) {
+      const node = this.map.get(key) as LinkedListNode<ExpiringSortedSetEntry<T>>
```
amit-momin, line 34: I think to avoid having to cast the object we could rearrange this check to something like this in all of these methods?
```suggestion
    const node = this.map.get(key)
    if (node) {
```

src/util/subscription-set/expiring-sorted-set.ts
```diff
@@ -16,28 +19,83 @@ interface ExpiringSortedSetEntry<T> {
...
+    return results
+  }
 
-    // Since we're iterating, might as well prune here
-    for (const [key, entry] of this.map.entries()) {
-      if (entry.expirationTimestamp < now) {
-        this.map.delete(key) // In theory, this shouldn't happen frequently for feeds
+  get(key: string): T | undefined {
+    if (this.map.has(key)) {
+      const node = this.map.get(key) as LinkedListNode<ExpiringSortedSetEntry<T>>
```
amit-momin, line 66: Same as top comment

src/util/subscription-set/expiring-sorted-set.ts
```diff
@@ -16,28 +19,83 @@ interface ExpiringSortedSetEntry<T> {
...
       }
+    } else {
+      return undefined
     }
+  }
 
-    return results
+  delete(key: string): void {
+    if (this.map.has(key)) {
+      const node = this.map.get(key) as LinkedListNode<ExpiringSortedSetEntry<T>>
```
amit-momin, line 81: Same as top comment

src/util/subscription-set/expiring-sorted-set.ts
```diff
@@ -16,28 +19,83 @@ interface ExpiringSortedSetEntry<T> {
...
+  }
+
+  private moveToTail(node: LinkedListNode) {
+    this.list.remove(node)
+    this.list.insertAtTail(node)
+  }
+
+  private evictIfNeeded() {
+    if (this.list.size >= this.capacity) {
+      logger.warn(`List reached maximum capacity, evicting least recently updated entry`)
```
amit-momin, line 94: Could we also log the item that's being removed here with the message? Think it could be useful when debugging.

src/util/subscription-set/expiring-sorted-set.ts
```diff
@@ -16,28 +19,83 @@ interface ExpiringSortedSetEntry<T> {
...
+  private evictIfNeeded() {
+    if (this.list.size >= this.capacity) {
+      logger.warn(`List reached maximum capacity, evicting least recently updated entry.`)
+      const node = this.list.removeHead()
+      if (node) {
+        logger.warn(`The node with key ${node.key} was removed`)
+        this.map.delete(node.key)
+      }
+    }
   }
```
amit-momin, line 100: This is just me nitpicking now but could you combine these 2 logs into one?  So there's no confusion if something gets printed in between? Something like: 
```suggestion
  private evictIfNeeded() {
    if (this.list.size >= this.capacity) {
      const node = this.list.removeHead()
      if (node) {
        logger.warn(`List reached maximum capacity, evicting least recently updated entry. The subscription with key ${node.key} was removed.`)
        this.map.delete(node.key)
      }
    }
  }
```

amit-momin (dismissed):

src/util/subscription-set/subscription-set.ts
```diff
@@ -9,27 +9,31 @@ import { RedisSubscriptionSet } from './redis-sorted-set'
  */
 export interface SubscriptionSet<T> {
   /** Add a new subscription to the set */
-  add(value: T, ttl: number): PromiseOrValue<void>
+  add(value: T, ttl: number, key?: string): PromiseOrValue<void>
```
amit-momin, line 12: Since you need the key for the expiring sorted set, I think you'll need to make a change in the abstract subscription transport to pass the cache key when we add to the subscription set

amit-momin: https://github.com/smartcontractkit/ea-framework-js/blob/063e3bb44a835871b7744c02e7dc9cc10c64cf2e/src/transports/abstract/subscription.ts#L44

PR #21, opened 2023-01-02
add desired params to message websocket handler

karen-stepanyan:
Currently only original message from DP is passed to `message` handler which makes it hard to construct normal responses where DP doesn't provide base and quote separately. 
For example a request with `{ base: 'ETH', quote: 'USD' } ` input params is sent to and received from DP as `symbol: ETHUSD` and without proper params it is not possible to construct cache entries. 
This PR passes all current subscriptions params to `message` handler.

### DIFF
--- a/src/transports/websocket.ts
+++ b/src/transports/websocket.ts
@@ -59,11 +59,13 @@ export interface WebSocketTransportConfig<T extends WebsocketTransportGenerics> 
      *
      * @param message - the message received by the WS
      * @param context - the background context for the Adapter
+     * @param params - current params in the subscription set
      * @returns a list of cache entries of adapter responses to set in the cache
      */
     message: (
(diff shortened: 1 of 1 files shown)


boxhock (approved):

amit-momin (approved):

alejoberardino:
There is a reason why this wasn't made, which is that due to the WS protocol being async (and any other async subscription transports for that matter, since if any changes were to be made they probably should go into those abstract classes) there will be a disconnect between the parameters and the responses sent by the server on its own timing. We cannot rely on the parameters we hold to be an accurate representation of what we receive after the fact. I understand the problem with providers replying with a combination of the original request parameters into the EA, but this is not a correct solution and should be reverted @boxhock @karen-stepanyan

PR #22, opened 2023-01-03
Feature/Replace cache key with stringified params

amit-momin:
- Replaced cache key generation implementation to use stringified input parameters instead
- Cleaned up subscription set implementation with new change in mind


src/cache/index.ts
```diff
@@ -112,45 +110,24 @@ export const calculateFeedId = <T extends EndpointGenerics>(
...
-        logger.debug(
-          `Property "${paramName}" in request parameters is of type object, and won't be used in the cacheKey`,
-        )
+  let cacheKey = JSON.stringify(data, (_, value) => {
+    if (value && typeof value === 'string') {
+      return value.toLowerCase()
     }
-  }
+    return value
+  })
```
boxhock, line 130: Does this mean that these inputs are now giving different cache keys?

A:

```
{
    "id": "1",
    "data": {
        "from": "ETH",
        "to": "USD",
        "overrides": {
            "coingecko": { "ETH": "ethereum"}
        }
    }
}
```

B:

```
{
    "id": "1",
    "data": {
        "from": "ETH",
        "to": "USD",
        "overrides": {
            "coingecko": { "ETH": "ethereum"},
            "coinpaprika": { "ETH": "ethereum"}
        }
    }
}
```

boxhock: If so, we need to block this PR and sit down and figure out how that is handled

amit-momin: If I understand correctly, I think both of these reduce down to something like these. I tested with an adapter to be sure.
`{"base":"eth","quote":"usd"} {"base":"ethereum","quote":"usd"}`

The input is rebuilt in the validator based on the input params so it would only contain those fields. Although if input params are array or object types, we could potentially generate different keys for the same input. But when I talked to Alejo about it, he wasn't concerned.

alejoberardino: The key reasoning as to why this would be safe lies in where this is being used: we will always generate the cache key after the data has been validated. Therefore the only thing that should be present for the stringify call should be valid input parameters, and that's a known, reduced subset without meta-parameters like overrides. It's true, however, that this is not accounting for unsorted arrays or nested structures, that's something easy to tackle but should be done carefully to make sure it's as performant as possible (i.e., if possible avoid sorting at all costs)

boxhock: Ok cool, since it's post-validator then it alleviates my original concern

src/cache/index.ts
```diff
@@ -112,45 +110,24 @@ export const calculateFeedId = <T extends EndpointGenerics>(
...
- * // equals `|base:eth|quote:btc`
+ * calculateKey({ base: 'ETH', quote: 'BTC' })
+ * // equals `{"base":"eth","quote":"btc"}`
  * ```
  */
 export const calculateKey = <CustomSettings extends SettingsMap>(
   data: unknown,
-  paramNames: string[],
   adapterConfig: AdapterConfig<CustomSettings>,
 ): string => {
```
boxhock, line 120: Need to update the JSDoc for this function. Seems like it was incomplete to begin with

boxhock (changes requested):
One small change, otherwise LGTM

PR #23, opened 2023-01-04
Treat literal "" env var values as empty

boxhock:


amit-momin (approved):

src/config/index.ts
```diff
@@ -397,7 +397,7 @@ const isEnvNameValid = (name: string) => /^[_a-z0-9]+$/i.test(name)
 export const getEnv = (name: string, config: Setting, prefix = ''): SettingValueType | null => {
   const value = process.env[getEnvName(name, prefix)]
 
-  if (!value) {
+  if (!value || value === '' || value === '""') {
```
alejoberardino, line 400: An empty string is already false, the middle condition is unnecessary

PR #28, opened 2023-01-05
Revert websocket message param change

amit-momin:


boxhock (approved):
Needs internal discussion, but approving for now to be on the safe side

PR #35, opened 2023-01-11
Feature/Consolidate metrics into class

amit-momin:
- Consolidated all metrics into a centralized class
- Initialize metrics only on adapter start


alejoberardino (commented):
We really shouldn't have to perform the assertion every time, I'd say to either trust that the metrics will be initialized and throw if they're not, or put them somewhere like the initialized dependencies and we pass them along

amit-momin:
> We really shouldn't have to perform the assertion every time, I'd say to either trust that the metrics will be initialized and throw if they're not, or put them somewhere like the initialized dependencies and we pass them along

Since I initialize only if metrics are enabled, I went with assertions so we could skip recording if metrics are disabled. But I could put it with initialized dependencies. I was considering it as an option earlier.

amit-momin:
> > We really shouldn't have to perform the assertion every time, I'd say to either trust that the metrics will be initialized and throw if they're not, or put them somewhere like the initialized dependencies and we pass them along
> 
> Since I initialize only if metrics are enabled, I went with assertions so we could skip recording if metrics are disabled. But I could put it with initialized dependencies. I was considering it as an option earlier.

Looking into it, I don't think dependencies are available everywhere we need to record metrics. Looks like trusting they're initialized is the way to go. But is that still a better approach than asserting and only recording when metrics are enabled?

boxhock:
Can we have the Metrics class initialized anyways, with a flag to enable metrics exporting `true/false`? If it's false, don't register the prom metrics. Then wrap the metric methods and check if metrics exporting is enabled or not?

It does solve that metric producers doesn't have to do the assertion every time, however an assertion is still done within the Metrics class then.

alejoberardino:
I think that in the future when we have a proper dependency injection framework/flow, dependencies would be the proper place for this, but it makes sense that now it becomes a pain: the current dependencies are more like "adapter dependencies" and note "code dependencies", if that makes sense.
I think that the current approach has some issues, mainly in practicality:
- Everything is "set", but the semantics don't match for every metric (many are counters, which should not be set but increased or decreased)
- Subtleties in typings are lost 
@amit-momin what do you think of doing something like this:
```typescript
class Metrics<T extends Record<string, unknown>> {
  private metrics?: T

  constructor(metricsDefinition: () => T) {
    if (process.env['METRICS_ENABLED'] === 'true') {
      this.metrics = metricsDefinition()
    }
  }

  get<K extends keyof T>(name: K): T[K] {
    const metric = this.metrics?.[name]
    if (!metric) {
      throw new Error(`Metric "${name as string}" was not initialized before use`)
    }
    return metric
  }
}

const transportMetrics = new Metrics(() => ({
  transportPollingDurationSeconds: new client.Gauge({
    name: 'transport_polling_duration_seconds',
    help: 'A histogram bucket of the distribution of transport polling idle time durations',
    labelNames: ['endpoint', 'succeeded'] as const,
  }),
  transportPollingFailureCount: new client.Counter({
    name: 'transport_polling_failure_count',
    help: 'The number of times the polling mechanism ran out of attempts and failed to return a response',
    labelNames: ['endpoint'] as const,
  }),
}))

transportMetrics.get('transportPollingDurationSeconds').inc()
```

Very crude PoC implementation, but the types should work. This has a couple of benefits:
- Barely any performance hit (an if statement), which is relevant when comparing alternatives to carry out the same behavior like [proxies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)
- Type safe, and no erasure on the different types of metrics (if you declare a Gauge metric, you'll get one, with all the methods that come with it)
- Very similar syntax to our existing usage, I'll admit that the function call to get the property doesn't look as pretty but the benefits outweigh the cons imo

amit-momin:
> @amit-momin what do you think of doing something like this:

Ya this is a much better idea. Another plus is we'd get to keep the metrics organized in their respective parts of the framework like before instead of this massive class.

boxhock:
@amit-momin You could just change it a bit to support that:

```
  get<K extends keyof T>(name: K): T[K] | undefined {
    if (!this.metrics) return // NOP
    const metric = this.metrics?.[name]
    if (!metric) {
      throw new Error(`Metric "${name as string}" was not initialized before use`)
    }
    return metric
  }
```

In use:

```typescript
transportMetrics.get('transportPollingDurationSeconds')?.inc()
```

amit-momin:
> @amit-momin You could just change it a bit to support that:

I was thinking about that but for histograms I'd probably need to assert for the timers if that's alright. Something like:
```
const metricsTimer = transportMetrics.get('transportPollingDurationSeconds')?.startTimer()

metricsTimer && metricsTimer()
```

amit-momin:
> The problem with assertions is that it would be really easy to forget to add them. Since it's very, very unlikely we'd ever run production instances without metrics, I'd suggest looking into returning a Proxy if the metrics have not been initialized

It may be simpler if I just initialize metrics regardless of the METRICS_ENABLED flag since that's the existing behavior. What do you think @boxhock

boxhock (approved):

PR #37, opened 2023-01-17
Add WebsocketTransport wrapper with reverse mapping

boxhock:


src/transports/websocket.ts
```diff
@@ -332,3 +332,18 @@ export class WebSocketTransport<
     }
   }
 }
+
+export class WebsocketReverseMappingTransport<
```
amit-momin, line 336: Could we add tsdocs for this to explain when this should be used?

boxhock: Should be covered now!

amit-momin (approved):

PR #43, opened 2023-01-20
Create New v3 EA guide

amit-momin:


amit-momin:
@smartcontractkit/ea-v3-engineers The doc is still in progress but wanted to open this up so you guys can review it in parts. Please let me know if I'm going into too much detail or not enough.

docs/porting-a-v2-ea-to-v3.md
```diff
@@ -276,6 +278,7 @@ export const adapter = new PriceAdapter({
     },
   },
   overrides: overrides['adapterName'], // If applicable, reference the overrides config
+  includes, // If applicable, reference the includes config
```
boxhock, line 281: 👍

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,655 @@
...
+    description: 'Exclusive from param1',
+  },
+} as const
+```
+
+### Cache Key Generator
+
+The `cacheKeyGenerator` method allows the developer to provide custom logic used to generate a unique cache key for each request. The framework's default behavior is to concat the endpoint name with the stringified input parameters such as `test-{"base":"ETH","quote": "USD"}`.
+
+A potential scenario of where a custom cache key generator would be used is if an input parameter is an array or an object. To ensure that a request generates the same unique key every time, the array or object would need to be sorted. An example of this code is shown below.
```
boxhock, line 278: I think, not just here but for the entire guide, we should make it clear when extra efforts should be made to make it non-standard (e.g. cache key generation). We should make it clear that developers should stick to the default unless they have to make it custom.

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,655 @@
...
+```typescript
+import { httpTransport } from './crypto-http'
+import { wsTransport } from './crypto-ws'
+
+export const routingTransport = new RoutingTransport<EndpointTypes>(
+  {
+    WS: wsTransport,
+    REST: httpTransport,
+  },
+  (req, adapterConfig) => (adapterConfig.WS_ENABLED ? 'WS' : 'REST'),
```
boxhock, line 353: While not ready yet, just want to flag that this will be different before end of February.

boxhock (commented):
I know it's still in progress, but I think you're going in the right direction 👍

docs/creating-a-new-v3-ea.md
amit-momin: Really good point. I'm offering up some of these things as features when I should be treating them as worst case scenarios.

docs/creating-a-new-v3-ea.md
amit-momin: I was thinking about that when adding this. For now, I'll add a `Subject to change` disclaimer

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,655 @@
...
+export const endpoint = new PriceEndpoint<EndpointTypes>({
+  ...
+})
+```
+
+## Transports
+
+A transport is the method in which the data is retrieved from a data provider for that particular endpoint. As seen in the [Endpoints](#endpoints) section, every endpoint requires one and only one transport. Define this transport in the same file as its associated endpoint. To learn more about specific transports, check the [Transport Types](#transport-types) section.
+
+### Routing Transport
```
alejoberardino, line 338: I don't think we should explain what a transport is here, but rather link to the basics doc and change this to be instead something like "Choosing a transport"

alejoberardino (changes requested):
Overall it looks good, but I would change the direction a bit to change this into more of a step by step guide to create a new EA and separate the _detailed_ description of what things are in the other place (a brief explanation is fine though, since some people might not even go read the other one)

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,371 @@
...
+├─ CHANGELOG.md
+├─ README.md
+├─ test-payload.json
+├─ tsconfig.json
+└─ tsconfig.test.json
+```
+
+## Adapter
+
+The adapter is initialized in the root-level `index.ts` file. Most of the parameters shown below for the Adapter are optional and only needed in particular cases. The parameters that are not self-explanatory are covered in their own sub-sections below.
```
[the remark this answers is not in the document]
amit-momin: Ok I'll keep this doc to the required components of an adapter and create separate docs for each component where I can go into more details. Like for adapter, endpoints, transports (already separated out), and tests? Just to keep it more bite-size.

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,371 @@
+# Creating a new v3 EA
+
+The v3 framework currently only supports development of source adapters. This guide will assume the adapter is being created in [External Adapters Monorepo](https://github.com/smartcontractkit/external-adapters-js) under packages/sources/
```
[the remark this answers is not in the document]
amit-momin: I made sure to add the distinction and also included a short section on "Types of adapters" but it's just definitions. Let me know if you want me to add some more detail.

docs/creating-a-new-v3-ea.md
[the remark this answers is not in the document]
amit-momin: I reorganized the way I mentioned above. Most of the content stayed the same in the individual component docs but just reworded to make them more standalone

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,371 @@
...
+      professional: {
+        rateLimit1s: 20,
+        rateLimit1m: 200,
+        rateLimit1h: 456.78,
+      },
+    },
+  },
+  customSettings, // If applicable, any custom settings
+  overrides: overrides['adapterName'], // If applicable, reference the overrides config.
+  envDefaultOverrides: { API_TIMEOUT: 10000 }, // If applicable, set overrides for env vars defaults. Primarily useful for framework level env vars.
```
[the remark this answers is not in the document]
amit-momin: Good point! I removed these details but kept some of the examples for things like `rateLimitConfig`, `overrides`, `includes` just for formatting in their sub-sections.

docs/creating-a-new-v3-ea.md
```diff
@@ -0,0 +1,371 @@
...
+        "inverse": true
+      }
+    ]
+  }
+]
+```
+
+### Request Transforms
+
+Request transforms is an array of functions that allow the transformation of input parameters. The functions are provided with the incoming `AdapterRequest` object that can be altered and passed on to the next function or the rest of the adapter's logic.
```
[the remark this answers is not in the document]
amit-momin: I added a diagram at the end of the Adapter doc for this. Realized the Basics doc had the lifecycle diagrammed broadly so I kept this one simple and just the order that validation and transformations occur before cache key generation

alejoberardino (changes requested):

alejoberardino (approved):

PR #52, opened 2023-01-26
Export ProviderRequestConfig & WebsocketTransportGenerics

boxhock:
We used to export ProviderRequestConfig previously, now we don't. This PR we don't don't export.

It helps typing EAs.

### DIFF
--- a/src/transports/http.ts
+++ b/src/transports/http.ts
@@ -37,8 +37,8 @@ type HttpTransportGenerics = TransportGenerics & {
 /**
  * Structure containing the association between EA params and a provider request.
  */
-type ProviderRequestConfig<T extends HttpTransportGenerics> = {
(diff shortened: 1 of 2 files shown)


amit-momin (approved):

amit-momin:
I feel like this change is more of a patch. Looks like the labels defaulted to feature/minor

boxhock:
@amit-momin Speaking strictly from Semver, a minor version bump would be the right thing, as we're adding types that consumers can use. A patch version change should be backwards compatible, which this is not. What do you think?

amit-momin:
> @amit-momin Speaking strictly from Semver, a minor version bump would be the right thing, as we're adding types that consumers can use. A patch version change should be backwards compatible, which this is not. What do you think?

You're right! Minor makes more sense. I'll change it back

PR #58, opened 2023-02-03
Add adapter name to cache key

amit-momin:
Added the adapter name to the cache key to avoid collisions between adapters when using Redis.


alejoberardino (approved):
Perfect, ty for the tests :smile:

PR #60, opened 2023-02-03
Add endpoint name to http request queue key

amit-momin:
Using only input parameters to generate a key for the request queue allowed requests from different transport to be coalesced. However, transports with the same input parameters could use different DP endpoints that yield different responses. Coalescing requests in this scenario would be incorrect and could lead to errors.

Added the endpoint name to the request queue key to allow coalescing only for requests from the same transport.


src/transports/http.ts
```diff
@@ -192,21 +192,25 @@ export class HttpTransport<T extends HttpTransportGenerics> extends Subscription
...
   ): Promise<TimestampedProviderResult<T>[]> {
     try {
       // The key generated here that we pass to the requester is potentially very long, but we're not considering it an issue given that:
       //   - the requester will store values in memory, so we're not sending the string anywhere
       //   - there's no problems using very large strings as object keys
       //   - there should be a limit on the amount of subscriptions in the set
+      // Append endpoint name to key to avoid coalescing requests across different endpoints
+      const requestKey = `${context.endpointName}-${requestConfig.params
+        .map((p) => JSON.stringify(p))
+        .join('|')}`
```
alejoberardino, line 205: If we're going to add the endpoint, would it make sense to just use the cache key?

alejoberardino (changes requested):

src/transports/http.ts
amit-momin: Yup that's a much better idea

alejoberardino (approved):

PR #63, opened 2023-02-08
Add transport input parameter for routing transport

alejoberardino:


src/transports/meta/routing.ts
```diff
@@ -1,30 +1,66 @@
...
 import { AdapterError } from '../../validation/error'
 import { MetaTransport, Transport, TransportDependencies, TransportGenerics } from '..'
+import { InputParameters } from '../../validation'
 
 const logger = makeLogger('RoutingTransport')
 
+export type RoutingTransportGenerics = TransportGenerics & {
+  Request: RequestGenerics & {
+    Params: RequestGenerics['Params'] & {
+      transport: string
```
amit-momin, line 14: This takes care of `transport` in the data object but I could see someone using it like endpoint and specifying it outside. Should we cover this scenario too?
```
{
   "endpoint": "price",
   "transport": "websocket",
   "data": {...}
}
```

alejoberardino: No, we discussed options and landed on specifying in the data payload so the cache key would discriminate between transports. I think adding ambiguity will lead to problems, we should reduce the scope as much as possible.

amit-momin: Good enough for me 👍

test/transports/routing.test.ts
```diff
@@ -456,6 +455,246 @@ test.serial('RoutingTransport can route to SSE transport', async (t) => {
...
+  const error: AxiosError | undefined = await t.throwsAsync(makeRequest)
+
+  t.is(error?.response?.status, 400)
+  t.is(
+    (error?.response?.data as { error: { message: string } }).error?.message,
+    'No transport found for qweqwe',
+  )
+})
+
+test.serial('missing transport in input params with no default fails request', async (t) => {
```
amit-momin, line 606: Not sure if I missed it but is there a test for missing transport but a default specified?

amit-momin: Oh sorry missing transport is probably the wrong phrasing. I meant if one isn't specified in the request but there's a default specified. Just another success scenario. Think all of the ones above have a transport specified.

amit-momin (approved):

PR #69, opened 2023-02-09
Modify workflows

alejoberardino:
This PR makes a few changes:
- Reorganizes label validation to run more consistently, publish comments and block PRs if the labels are missing or incorrect
- Refactors usage of the `::set-output` directive that is being deprecated
- Move publishing to its own workflow, changing the trigger from a closing version-bump PR to a push to main


.github/workflows/add-or-validate-labels.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - opened
+      - labeled
+      - unlabeled
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  add-labels: # Load the dependencies cache for all jobs
```
amit-momin, line 17: Since this is consolidated now, I assume we want to run this not only on open anymore. This is where my lack of knowledge on GH workflows kick in. Would this cause any conflicts/issues if we manually update the labels to something that's different than the calculation? Like if the branch name was incorrectly named.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,38 @@
...
+    outputs:
+      VERSION_TAG: ${{ steps.get-version-tag.outputs.VERSION_TAG }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/setup
+      - run: yarn build
+      - run: |
+          # Can't use `npm show . version` because that will show the version of the package in the registry, not the version in package.json, and we haven't published yet
+
+          params=(--access restricted)
```
amit-momin, line 28: I think this should be `public` now right?
```suggestion
          params=(--access public)
```

alejoberardino: You're right, fixed

.github/workflows/add-or-validate-labels.yaml
alejoberardino: No, this specific job that adds labels we do want to open only when the PR is opened. After that we can just listen for the label changes. And for the second part, if the user then changes the labels that were auto-added, the validation should still work as expected, and no conflicts should arise

amit-momin: I see. I very conveniently missed line 20

amit-momin (approved):

PR #75, opened 2023-02-22
remove smoke endpoint

karen-stepanyan:
This PR
 removes `buildSmokeEndpoint` function from src/index.ts
removes `SMOKE_TEST_PAYLOAD_FILE_NAME` config
removes test-payload-loader file 
removes all smoke tests


alejoberardino (changes requested):
Let's remove all mentions from the docs as well

alejoberardino (approved):

PR #78, opened 2023-02-23
Remove nock, bump axios, fix error logging

alejoberardino:
`nock` is very easy to use, but it has one problem given our suite of tests: most of them also mock the global clock, which nock also relies on internally (uses `nextTick` to make things more "realistic", for example). This PR removes nock and uses `axios-mock-adapter` instead, which is a more straightforward approach (just an axios adapter that matches request and response). The drawback is that we won't be mocking _every_ network call in the adapter, but nock wasn't foolproof either and in theory we'd only use axios to make the internal http calls anyways.
Additionally, this PR bumps axios to the latest stable version and fixes the error logging in the error catching middleware.


test/metrics/metrics.test.ts
```diff
@@ -146,7 +154,7 @@ test.serial('Test http request duration metrics', async (t) => {
 test.serial('Test data provider requests metrics', async (t) => {
   const metricsMap = await t.context.testAdapter.getMetrics()
   const expectedLabel = `{provider_status_code="200",method="GET",app_name="TEST",app_version="${version}"}`
-  t.is(metricsMap.get(`data_provider_requests${expectedLabel}`), 1)
+  t.is(metricsMap.get(`data_provider_requests${expectedLabel}`), 9)
```
amit-momin, line 157: I may have missed something but what's causing these metrics to have a jump in their numbers?

alejoberardino: Because this test didn't use to use a mocked system clock, and now that it does in the methods we use in the testing classes to wait for results we also account for the background executes. That makes it so each time we gradually advance the clock by iterating over the scheduled async calls, we could be firing more background executes in the meantime

amit-momin: Oh ok I see. So we are actually sending more than just the one request because of the background execute.

amit-momin (approved):

PR #82, opened 2023-02-24
Remove defaultTransport parameter from RoutingTransport

alejoberardino:
The default transport shouldn't be set in the routing transport class, because it would cause problems with cache keys. The default should be set in the input parameters instead.


docs/v3-ea-components/transport-types/routing-transport.md
```diff
@@ -25,8 +28,6 @@ export const routingTransport = new RoutingTransport<EndpointTypes>(
     ws: wsTransport,
     rest: httpTransport,
   },
-  {
```
amit-momin, line 28: Doesn't let me add a comment lower but think a "}" needs to be removed too

amit-momin (approved):

PR #83, opened 2023-02-26
Add background execute errors metric, transport label, more

alejoberardino:
This PR covers a lot of ground:
- Update AVA
- Add background execute errors metric
- Add transport label to background execute metrics
- Improve rate limiting logging
- Heavily improve requester logging
- Fix issue where logging context would be shared between logs on the same async thread
- Add test cases
- Revamp metrics testing, with a cleaner and easier to use interface, and very helpful runtime assistance


src/util/requester.ts
```diff
@@ -237,17 +237,26 @@ export class Requester {
...
     } catch (e) {
+      const err = e as AxiosError
+      logger.info({
+        msg: 'Request failed',
+        response: {
+          statusCode: err.response?.status,
+          data: err.response?.data,
+          text: err.response?.statusText,
+        },
+      })
```
amit-momin, line 248: Was this intentionally marked as `info` since this is a data provider error?

alejoberardino: Correct

amit-momin (approved):

PR #85, opened 2023-02-27
deep validation for input objects and arrays

karen-stepanyan:
[PDI-2025](https://smartcontract-it.atlassian.net/browse/PDI-2025)

[PDI-2025]: https://smartcontract-it.atlassian.net/browse/PDI-2025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/validation/input-validator.ts
+++ b/src/validation/input-validator.ts
@@ -9,6 +9,8 @@ export type Override = Record<string, string>
 export type OverridesMap = Record<string, Override>
 
 const isNotObject = validator.object()
+const MAX_ARRAY_LENGTH = 20
+const MAX_OBJECT_SIZE = 20
 
 export class InputValidator {
   private aliases: { [key: string]: string[] } = {}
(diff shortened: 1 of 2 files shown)


src/validation/input-validator.ts
```diff
@@ -140,26 +142,62 @@ export class InputValidator {
...
+    }
+
+    data.forEach((item, index) => {
+      if (!['string', 'number', 'boolean'].includes(typeof item)) {
+        this.throwInvalid(
+          `${key} parameter must be an array of strings|numbers|booleans, received '${typeof item}' at index ${index}`,
+        )
+      }
+    })
+  }
```
amit-momin, line 175: I could see these validations causing issues for PoR adapters on v3. Depending on the pool of addresses for a client, we could have an array with thousands of addresses. And with the current format, it would be an array of objects. Maybe we could make max items configurable and also allow objects? Something to maybe discuss with Alejo. An example is something like below.
```
{
    "data": {
        "addresses": [
            ...
            {
                "address": "0x0000"
            },
            ...
        ]
    }
}
```

karen-stepanyan: @alejoberardino any thoughts on this ?

alejoberardino:
This was covered instead in the new input params impl

PR #86, opened 2023-02-28
Refactor transport routing

alejoberardino:
There was a fundamental problem with how the RoutingTransport was structured. When we decided to put the transport in the input params, there was a conceptual clash that wasn't that apparent: the transport indication coming in the request is a meta-parameter, not really relevant for transports themselves but more for altering the behavior (routing) of the request through the adapter. This meant that cases like a default transport would cause issues in many ways, according to how it was implemented:

- If the default was set in the input parameters, then custom routers could not properly divert to any non-default transport due to the params cache key generated being different from the eventual transport that would fetch the data.
- If the default was set in the routing transport, there was also no way to edit the params, and the cache key would also be different but when the transport wasn't specified and the default wasn't set.

Because of this, this PR changes the routing concept in the framework. Summarized:
- Request transforms have been moved to the adapter endpoints.
- Routing has been moved from a RoutingTransport to become part of the AdapterEndpoint.
- Cache keys now include the transport as part of the string.


src/adapter/price.ts
```diff
@@ -69,6 +69,8 @@ type IncludesMap = Record<string, Record<string, IncludeDetails>>
  * Data Feeds, its InputParameters must extend the basic ones (base/quote).
  */
 export class PriceEndpoint<T extends PriceEndpointGenerics> extends AdapterEndpoint<T> {
+  includesMap?: IncludesMap
```
amit-momin, line 72: Noticed you kept the logic for includesMap in the PriceAdapter. Is this needed?

alejoberardino: removed

src/adapter/endpoint.ts
```diff
@@ -1,40 +1,75 @@
...
+export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpointInterface<T> {
   name: string
   aliases?: string[] | undefined
-  transport: Transport<T> | MetaTransport<T>
+  transports: Record<string, Transport<T>>
   inputParameters: SpecificInputParameters<T['Request']['Params']>
   rateLimiting?: EndpointRateLimitingConfig | undefined
   validator: InputValidator
   cacheKeyGenerator?: (data: Record<string, unknown>) => string
   customInputValidation?: CustomInputValidator<T>
```
amit-momin, line 33: Now that the transport field isn't in the input params, `customInputValidation` can't validate which transport the request is going to since its parameters are input params and config. Think we may need to update it to have transportName that we can pass in the validator. We need it for adapters that have optional WS transports to be able to throw an error if a request comes in for ws when the api key isn't set.

amit-momin: Noticed this when I was trying the new changes on `tradermade-test`

alejoberardino: Valid point, I've given the custom input validator the entire request, don't really love the patterns that could come of it, but it's going to solve many problems in the short term

amit-momin (approved):

PR #90, opened 2023-03-03
Refactor Settings

alejoberardino:
The current implementation of custom settings caused a lot of issues with generics. The problem was mainly around the fact that the type for the resolved settings (that is, the key:value pair once processed) was also itself a generic parameter, and it really messed up with polymorphism. Things like 
```typescript
const adapter = new Adapter({
	name: 'TEST',
	endpoints: [customEndpoint] // Endpoint is of type CustomEndpoint, with typeof customSettings
	customSettings: customSettings
})
```
Would fail in the assignment of the endpoint to the array of endpoints for the adapter parameters, due to a clash in settings. Another thing was a bit annoying, in that in order to define the settings in a way that added type safety one needed to make use of the `satisfies` keyword. There is nothing necessarily wrong with that, but it's a newer feature and developers might not be very familiar with it. 

This PR refactors the way the adapter config is defined, providing a class based approach that allows for cleaner generics and a type safe way to define them without asking a lot of the developer.

### DIFF
--- a/src/adapter/types.ts
+++ b/src/adapter/types.ts
@@ -105,7 +105,7 @@ export interface AdapterParams<T extends AdapterConfig = AdapterConfig> {
   /** Bootstrap function that will run when initializing the adapter */
   bootstrap?: (adapter: Adapter<T>) => Promise<void>
 
-  /** TODO: complete */
+  /** The custom [[AdapterConfig]] to use. If not provided, the default configuration will be used instead */
   config?: T
 }
 
--- a/src/adapter/types.ts
+++ b/src/adapter/types.ts
@@ -1,7 +1,7 @@
 import type EventSource from 'eventsource'
 import Redis from 'ioredis'
 import { Cache } from '../cache'
-import { BaseAdapterConfig, ProcessedConfig, SettingsMap } from '../config'
+import { BaseAdapterSettings, ProcessedConfig, SettingsDefinitionMap } from '../config'
 import { AdapterRateLimitTier, RateLimiter } from '../rate-limiting'
 import { Transport, TransportGenerics } from '../transports'
 import { AdapterRequest, SingleNumberResultResponse, SubscriptionSetFactory } from '../util'
@@ -11,9 +11,9 @@ import { AdapterError } from '../validation/error'
 import { Adapter } from './basic'
 import { AdapterEndpoint } from './endpoint'
 
-export type CustomAdapterSettings = SettingsMap & NegatedAdapterSettings
+export type CustomAdapterSettings = SettingsDefinitionMap & NegatedAdapterSettings
 type NegatedAdapterSettings = {
-  [K in keyof BaseAdapterConfig]?: never
+  [K in keyof BaseAdapterSettings]?: never
 }
 
 /**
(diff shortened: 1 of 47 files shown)


src/adapter/types.ts
```diff
@@ -99,17 +99,14 @@ export interface AdapterParams<CustomSettings extends SettingsMap> {
...
-  customSettings?: SettingsMap
-
   /** Configuration relevant to outbound (EA --\> DP) communication rate limiting */
   rateLimiting?: AdapterRateLimitingConfig
 
   /** Bootstrap function that will run when initializing the adapter */
-  bootstrap?: (adapter: Adapter<CustomSettings>) => Promise<void>
+  bootstrap?: (adapter: Adapter<T>) => Promise<void>
+
+  /** TODO: complete */
```
austinborn, line 108: TODO - can you either finish this or add a ticket for it (if there isn't one already)?

src/config/index.ts
```diff
@@ -519,11 +506,98 @@ export type AdapterConfigFromSettings<T extends SettingsMap> = {
...
+export type EmptySettingsDefinitionMap = Record<string, never>
+export type SettingsDefinitionMap = Record<string, SettingDefinition>
 export type ValidationErrorMessage = string | undefined
+
+/**
+ * TODO: Improve docs
+ * This class will hold the processed config type, and the basic settings
+ * The idea is that you can no longer use a straight object, but have to build a config
+ * Then in the generics we can simply pass the type of this, and it will hopefully allow for simpler generics
+ * name is placeholder
```
austinborn, line 523: Name still placeholder?

src/config/index.ts
```diff
@@ -519,11 +506,98 @@ export type AdapterConfigFromSettings<T extends SettingsMap> = {
...
+export class AdapterConfig<T extends SettingsDefinitionMap = SettingsDefinitionMap> {
+  settings!: AdapterSettings<T>
+
+  constructor(
+    private settingsDefinition: T,
+    private options?: {
+      /** Map of overrides to the default config values for an Adapter */
+      envDefaultOverrides?: Partial<BaseAdapterSettings>
+
+      /** TODO: complete */
```
austinborn, line 534: Another TODO

src/config/index.ts
```diff
@@ -519,11 +506,98 @@ export type AdapterConfigFromSettings<T extends SettingsMap> = {
...
+          setting.sensitive &&
+          (this.settings as Record<string, ValidSettingValue>)[name],
+      )
+      .map(([name]) => ({
+        key: name,
+        // Escaping potential special characters in values before creating regex
+        value: new RegExp(
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          ((this.settings as AdapterSettings)[name]! as string).replace(
+            /[-[\]{}()*+?.,\\^$|#\s]/g,
```
austinborn, line 589: Did you come up with this or pull it from somewhere?

austinborn (approved):
Looks fine, just a couple open TODOs that should be resolved or documented. Nice work!

src/adapter/types.ts
alejoberardino: Completed

src/config/index.ts
alejoberardino: No, I had already renamed 😄

src/config/index.ts
alejoberardino: Completed

src/config/index.ts
alejoberardino: Not me I think, I moved this from somewhere else in the code

austinborn (approved):

PR #95, opened 2023-03-06
add docs for http transport

karen-stepanyan:


docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+            result: result.price,  
+          },  
+          result: result.price,  
+        },  
+      }  
+    })  
+  },  
+})
+```
+### Building requests
```
boxhock, line 37: Another nitpick.

I can't do a suggested change because the formatting gets messed up, but we should have one open line between L36-L37 and L37-L38.

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+**prepareRequests**  method  takes the list of currently active input parameters in the subscription set and adapter config and should return one or multiple request configs (_ProviderRequestConfig_) depending if Data Provider supports request batching or not.
```
boxhock, line 38: Trying to make this a bit more clear:

```suggestion
The **prepareRequests**  method  takes the list of currently active input parameters in the subscription set and the adapter config. If the Data Provider only supports requesting a single data point for each HTTP request, the method should return a single request config (_ProviderRequestConfig_). However, if the Data Provider supports requesting multiple data points in each HTTP request, the method should return an array of request configs.
```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+
+#### Request config structure
+*ProviderRequestConfig* is a object that contains two keys `params` and `request` . `params` is the input parameters for a request and whatever  is provided will be available in the **parseResponse** method. `request` is the *axios* request configuration where request url, request method, query or body params, headers and etc can be specified.
```
boxhock, line 41: > `params` is the EA input parameters for a request

This^ is accurate, right?

```suggestion
*ProviderRequestConfig* is an object with two keys: `params` and `request`. `params` is the EA input parameters for a request, and will also be available in the **parseResponse** method. `request` is the *axios* request configuration where parameters such as the request URL, request method, query or body params, and headers can be specified.
```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+
+#### Send batch request to Data Provider
+The example above assumes that Data Provider doesn't support batching and each active input parameter will initiate request to Data Provider. Bellow is the example of batch version of prepareRequests
+```typescript
```
boxhock, line 45: More spacing

```suggestion
#### Send batch request to Data Provider

The example above assumes that the Data Provider doesn't support batching, and each active input parameter will initiate a request to Data Provider. Below is an example of a batch version of prepareRequests.

```typescript  
```

boxhock: > The example above assumes

Which example is it referring to?

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+          baseURL: config.API_ENDPOINT,  
+          url: '/price',
+          params: {
+	          symbols: [...new Set(params.map((p) => p.symbol.toLowerCase()))].join(',')
+          }  
+        },  
+      } 
+  }
+```
+The difference is that instead of returning array of request config objects, only one object is returned where information about input parameters is passed to Data Provider as one query parameter.
```
boxhock, line 59: Spacing

```suggestion

The difference is that instead of returning array of request config objects, only one object is returned where information about input parameters is passed to Data Provider as one query parameter.
```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+        params: param, // Note that this should be single input parameter, not the params array above
+        response: {  
+          data: {  
+            result: value,  
+          },  
+          result: value,  
+        },  
+      }  
+    })  
+  }
```
boxhock, line 113: Indentation is off + need spacing here as well

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+```
+### Type params of the transport
+`HttpTransport` accepts one type param that describes the request and response types for the endpoint and the transport. It is used to ensure that the transport and endpoint are compatible.
+```typescript
+const transport = new HttpTransport<EndpointTypes>({
+```
+Example structure of `EndpointTypes`
+```typescript
```
boxhock, line 121: Should all be one block?

boxhock (changes requested):

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+export type EndpointTypes = {  
+ Request: {  
+	Params: CryptoRequestParams // type of input parameters
+  }  
+ Response: SingleNumberResultResponse // type of External Adapte response. `SingleNumberResultResponse` is built in type that indicates that both `data` and `result` are numbers
+ Settings: BaseAdapterSettings  // type of adapter config
+ Provider: {  
+  RequestBody: never  // type of request body in case of POST requests. `never` for GET requests
```
boxhock, line 129: You should technically be able to have a request body in a GET request, but usually not. Just a nit pick

```suggestion
  RequestBody: never  // type of request body in the case of POST requests. This is usually `never` for GET requests, and is *not* the same as query params
```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+### Type params of the transport
+`HttpTransport` accepts one type param that describes the request and response types for the endpoint and the transport. It is used to ensure that the transport and endpoint are compatible.
+```typescript
+const transport = new HttpTransport<EndpointTypes>({
+```
+Example structure of `EndpointTypes`
+```typescript
+export type EndpointTypes = {  
+ Request: {  
+	Params: CryptoRequestParams // type of input parameters
```
boxhock, line 124:
```suggestion
	Params: CryptoRequestParams // type for the EA endpoint input parameters
```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,135 @@
...
+  }  
+ Response: SingleNumberResultResponse // type of External Adapte response. `SingleNumberResultResponse` is built in type that indicates that both `data` and `result` are numbers
```
boxhock, line 126:
```suggestion
 Response: SingleNumberResultResponse // type of External Adapter response. `SingleNumberResultResponse` is a built in type that indicates that both `data` and `result` are numbers
```

boxhock (changes requested):

docs/v3-ea-components/transport-types/http-transport.md
karen-stepanyan: regarding `params` i would say yes and. no :). it will be more accurate to say
 `params` is the non-expired input parameters in the subscription set, but i think it will be more confusing or somewhere we need to also mention what is that subscription set and how it works.

docs/v3-ea-components/transport-types/http-transport.md
karen-stepanyan: the very first `Example usage of HttpTransport` .  maybe we can rename it like that ? 'the first example' ?

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,153 @@
...
+ Response: SingleNumberResultResponse // type of External Adapter response. `SingleNumberResultResponse` is a built in type that indicates that both `data` and `result` are numbers
+ Settings: BaseAdapterSettings  // type of adapter config
+ Provider: { 
+  RequestBody: never  // type of request body in the case of POST requests. This is usually `never` for GET requests, and is *not* the same as query params
+  ResponseBody: ProviderResponseBody  // interface of raw response body from Data Provider 
+  }  
+}
+```
+
+
```
boxhock, line 153:
```suggestion

```

docs/v3-ea-components/transport-types/http-transport.md
```diff
@@ -1,3 +1,153 @@
...
+ Request: {  
+	Params: CryptoRequestParams // type for the EA endpoint input parameters
+  }  
+ Response: SingleNumberResultResponse // type of External Adapter response. `SingleNumberResultResponse` is a built in type that indicates that both `data` and `result` are numbers
+ Settings: BaseAdapterSettings  // type of adapter config
+ Provider: { 
+  RequestBody: never  // type of request body in the case of POST requests. This is usually `never` for GET requests, and is *not* the same as query params
+  ResponseBody: ProviderResponseBody  // interface of raw response body from Data Provider 
+  }  
+}
```
boxhock, line 150: Nitpick, but we're mixing tabs/spaces here, and there's some trailing whitespace on some lines

boxhock (changes requested):

docs/v3-ea-components/transport-types/http-transport.md
boxhock: Ok, I'm fine with how it is now!

boxhock (dismissed):

alejoberardino (approved):

PR #96, opened 2023-03-06
Add background execute and websocket open timeouts

alejoberardino:
This PR attempts to mitigate issues found with v3 EAs, primarily websocket ones (DAR, Galaxy). Those EAs work perfectly for a period of time, then suddenly we can observe the background executions drop to zero. This means one of two things:

- **The background execution loop is crashing.** This seems very unlikely to be the culprit because there's no logs to this end, and one would expect them (things like unhandled promises). I have verified that some output would be present by simply throwing an exception manually from the loop, and it results in both a very clear log message and the process exiting.
- **The background execution loop is stuck.** This is the current best guess. If an issue were happening in the background execution, looking at the current logic that is in the deployed adapters it could be possible for a bad execute to hang forever.

In an effort to both verify that the current hypothesis is true and mitigate it, this PR includes the following:

- Adds logic to time out background executes. It is set to a configurable time of 90s, but this should be more than enough to cover all current execution timings (in the future it can be dialed down).
- Same logic applied to the WebSocket open handler. This one is set to 10s, also configurable.
- Some rearranging of the WS opening logic so that if we have this stuck connection, it won't affect the overall loop and just stay in the background.
- Addition of a WS closure metric, including code, to better track the causes of WS closures from DPs.

A note on the way the timeouts are implemented: it's not straightforward to "cancel" the execution of an async promise. The best course of action seems to be to try and guarantee, as best as possible, that the promise that times out won't have any side effects if it eventually resolves. This of course should raise concerns on the garbage collection of these promises, and the potential overflow they could cause. It is something I am aware of; for now it seems a safe approach due to the easy to spot nature of the memory increase (which from my testing is very very low, I'd need a very long running time to verify if it even happens at all).

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -90,3 +90,35 @@ export class DoubleLinkedList {
     return this.remove(this.head)
   }
 }
+
+class PromiseTimeoutError extends Error {}
+
+/**
+ * Waits for the provided promise to finish up to a specified amount of time, at which point
+ * an error is thrown if the promise hasn't finished yet. Note that this cannot abort the execution
+ * of the provided promise, as the underlying node structure cannot cancel the thread.
+ *
+ * @param promise - the promise to wait for
+ * @param timeout - the maximum amount of time to wait for the promise to finish
+ * @returns the result of the promise
+ */
+export const timeoutPromise = <T>(
+  label: string,
+  promise: Promise<T>,
+  timeout: number,
+): Promise<T> => {
+  let timer: NodeJS.Timeout
+  return Promise.race([
+    promise,
+    new Promise<T>((_, reject) => {
+      timer = setTimeout(
+        () =>
+          reject(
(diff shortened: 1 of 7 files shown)


src/util/index.ts
```diff
@@ -90,3 +90,35 @@ export class DoubleLinkedList {
     return this.remove(this.head)
   }
 }
+
+class PromiseTimeoutError extends Error {}
+
+/**
+ * Waits for the provided promise to finish up to a specified amount of time, at which point
+ * an error is thrown if the promise hasn't finished yet. Note that this cannot abort the execution
+ * of the provided promise, as the underlying node structure cannot cancel the thread.
```
boxhock, line 99: Time for a Go rewrite? :smile:

boxhock (approved):
Changes look good, with the caveat of cancelling promises in node, but this seems acceptable in the effort to mitigate the issues we've seen.

PR #108, opened 2023-03-14
Re add rate limiting strategies, fix tests & simple counting

alejoberardino:
The existing simple counting rate limiter is great to both respect the rate limits set by the config, while allowing quick bursts. Unfortunately for many providers, these bursts seem to trigger limits on their end. This PR reintroduces different limiting strategies, setting a "fixed" one as the default (basically, requests will be fired one by one, always waiting the same amount between each of them.
Also in this PR some GH workflows are fixed

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -330,7 +330,7 @@ export const BaseSettingsDefinition = {
     description: 'The rate limiting strategy to use for outbound requests',
     type: 'enum',
     options: ['counting', 'fixed'],
-    default: 'counting',
+    default: 'fixed',
   },
 } as const satisfies SettingsDefinitionMap
 
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -329,8 +329,8 @@ export const BaseSettingsDefinition = {
   RATE_LIMITING_STRATEGY: {
     description: 'The rate limiting strategy to use for outbound requests',
     type: 'enum',
-    options: ['counting', 'fixed'],
-    default: 'fixed',
+    options: ['burst', 'fixed-interval'],
+    default: 'fixed-interval',
   },
 } as const satisfies SettingsDefinitionMap
 
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -326,6 +326,12 @@ export const BaseSettingsDefinition = {
     default: 90_000,
     validate: validator.integer({ min: 1000, max: 180_000 }),
   },
+  RATE_LIMITING_STRATEGY: {
+    description: 'The rate limiting strategy to use for outbound requests',
+    type: 'enum',
+    options: ['counting', 'fixed'],
+    default: 'counting',
+  },
 } as const satisfies SettingsDefinitionMap
 
 export const buildAdapterSettings = <
--- a/src/rate-limiting/fixed.ts
+++ b/src/rate-limiting/fixed.ts
@@ -45,7 +45,6 @@ export class FixedRateLimiter implements RateLimiter {
       timeSinceLastRequest: ${timeSinceLastRequest}
       period: ${this.period}
       remainingTime: ${remainingTime}
-      remainingTime: ${remainingTime}
     `)
     return remainingTime
   }
--- a/src/rate-limiting/fixed.ts
+++ b/src/rate-limiting/fixed.ts
@@ -40,15 +40,14 @@ export class FixedRateLimiter implements RateLimiter {
 
     const timeSinceLastRequest = now - this.lastRequestAt // Positive int
     const remainingTime = Math.max(0, this.period - timeSinceLastRequest)
-    const timeToWait = Math.min(this.period, remainingTime)
     logger.trace(`Rate limiting details:
       now: ${now}
       timeSinceLastRequest: ${timeSinceLastRequest}
       period: ${this.period}
       remainingTime: ${remainingTime}
-      timeToWait: ${timeToWait}
+      remainingTime: ${remainingTime}
     `)
-    return timeToWait
+    return remainingTime
   }
 
   async waitForRateLimit(): Promise<void> {
--- a/src/rate-limiting/fixed.ts
+++ b/src/rate-limiting/fixed.ts
@@ -1,0 +1,64 @@
+import { AdapterRateLimitTier, RateLimiter } from '.'
+import { AdapterEndpoint, EndpointGenerics } from '../adapter'
+import { makeLogger, sleep } from '../util'
+
+const logger = makeLogger('FixedRateLimiter')
+
+/**
+ * The simplest version of a rate limit. This will not take any bursts into accoung,
+ * and always rely on a fixed request per second rate. The only "complex" mechanism it has
+ * is checking when the last request was made to this rate limiter, to account for a period
+ * of time with no requests and avoiding the wait of the initial request.
+ */
+export class FixedRateLimiter implements RateLimiter {
+  period!: number
+  lastRequestAt: number | null = null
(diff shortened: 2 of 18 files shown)


src/rate-limiting/fixed.ts
```diff
@@ -0,0 +1,64 @@
...
+ * and always rely on a fixed request per second rate. The only "complex" mechanism it has
+ * is checking when the last request was made to this rate limiter, to account for a period
+ * of time with no requests and avoiding the wait of the initial request.
+ */
+export class FixedRateLimiter implements RateLimiter {
+  period!: number
+  lastRequestAt: number | null = null
+
+  initialize<T extends EndpointGenerics>(
+    endpoints: AdapterEndpoint<T>[],
```
amit-momin, line 18: Do we still need endpoints in here? Looks like neither rate limiter uses it during initialization

src/rate-limiting/simple-counting.ts
```diff
@@ -65,24 +60,51 @@ export class SimpleCountingRateLimiter implements RateLimiter {
...
+    }
+  }
+
+  msUntilNextExecution(): number {
+    // If the limit is set to infinity, there was no tier limit specified
+    if (this.perSecondLimit === Infinity && this.perMinuteLimit === Infinity) {
+      return 0
+    }
+
+    const { now, nextSecondInterval, nextMinuteInterval } = this.updateIntervals()
```
amit-momin, line 76: I couldn't put the comment further up but weren't we going to consolidate the simple counting rate limits to just seconds? Because of scenarios like `RATE_LIMIT_CAPACITY_SECOND: 10` and `RATE_LIMIT_CAPACITY_MINUTE: 50`. We'd blow the credits within the first 5 seconds and have to wait till the next minute causing data to be stale over the next 55s.

src/rate-limiting/simple-counting.ts
```diff
@@ -65,24 +60,51 @@ export class SimpleCountingRateLimiter implements RateLimiter {
...
+
+  async waitForRateLimit(): Promise<void> {
+    const timeToWait = this.msUntilNextExecution()
+
     if (timeToWait === 0) {
-      this.requestsThisSecond++
-      this.requestsThisMinute++
       logger.trace(
-        `Request under limits, counted +1 (S = ${this.requestsThisSecond} | M = ${this.requestsThisMinute})`,
+        `Request under limits, current count: (S = ${this.requestsThisSecond} | M = ${this.requestsThisMinute})`,
```
amit-momin, line 92: Wouldn't this log be a little redundant with the last one? Since the next one implies the request is under limits?

src/rate-limiting/fixed.ts
alejoberardino: Not in the current state, but there's code to allow for endpoint allocation that we might want to use in the future, if we want to clean it up I'd say we do that later

src/rate-limiting/simple-counting.ts
alejoberardino: That was an alternative to the fixed rate limiting, with the different statuses it's ok if that's what happens with the counting strategy. In reality, this strategy should be used when the rate limiting on the provider comfortably allows for the regular load on the adapter, and the bursts would cover a random flurry of messages. I do agree that there might be "smarter" ways of handling it, but those should be implemented as separate strategies imo

src/rate-limiting/fixed.ts
amit-momin: Fair enough

src/rate-limiting/simple-counting.ts
alejoberardino: I do like to see current -> update, even if redundant. I know the code will be thread safe cuz JS, but my brain likes to have it because it looks for concurrency problems by default haha

src/rate-limiting/simple-counting.ts
amit-momin: If the adapter config allows regular load and bursts, I think consolidating to seconds would still allow for bursts to be used effectively. It'd just cover the scenarios that the longer term tier is lacking a little so we'd need smaller bursts over time. Incase someone accidentally configures `counting` without realizing how it works under the hood, consolidating feels safer. But I understand if we want to keep the large bursts up front so I'll leave it up to you if you want to resolve this comment.

src/rate-limiting/simple-counting.ts
amit-momin: I'm good with that 👍

amit-momin (approved):

src/rate-limiting/fixed.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (!this.lastRequestAt) {
+      logger.trace(
+        `First request for the rate limiter, sending immediately. All subsequent requests will wait ${this.period}ms.`,
+      )
+      return 0
+    }
+
+    const timeSinceLastRequest = now - this.lastRequestAt // Positive int
+    const remainingTime = Math.max(0, this.period - timeSinceLastRequest)
+    const timeToWait = Math.min(this.period, remainingTime)
```
amit-momin, line 43: Why do we have this variable? Wouldn't it always equate `remainingTime` since it's always less than or equal to `this.period`?

alejoberardino: You're right, I had this when the logic was a bit different. I'll remove

alejoberardino: You're right, I had this when the logic was a bit different. I'll remove

src/rate-limiting/fixed.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const timeSinceLastRequest = now - this.lastRequestAt // Positive int
+    const remainingTime = Math.max(0, this.period - timeSinceLastRequest)
+    logger.trace(`Rate limiting details:
+      now: ${now}
+      timeSinceLastRequest: ${timeSinceLastRequest}
+      period: ${this.period}
+      remainingTime: ${remainingTime}
+      remainingTime: ${remainingTime}
+    `)
```
amit-momin, line 49: Think `remainingTime` got duplicated in the log
```suggestion
    logger.trace(`Rate limiting details:
      now: ${now}
      timeSinceLastRequest: ${timeSinceLastRequest}
      period: ${this.period}
      remainingTime: ${remainingTime}
    `)
```

src/config/index.ts
```diff
@@ -326,6 +326,12 @@ export const BaseSettingsDefinition = {
     default: 90_000,
     validate: validator.integer({ min: 1000, max: 180_000 }),
   },
+  RATE_LIMITING_STRATEGY: {
+    description: 'The rate limiting strategy to use for outbound requests',
+    type: 'enum',
+    options: ['counting', 'fixed'],
+    default: 'fixed',
```
boxhock, line 333: This is something _additional_ documentation would help fix, but is there any way we can get more descriptive names of the strategies? I'd have no idea what the difference is between "counting" and "fixed" if I were to configure this.

alejoberardino: I agree that it's not super clear, but I couldn't think of any alternatives, any suggestions are welcome

amit-momin: It doesn't necessarily describe the strategy and more so the behavior of requests but what if we do `burst` and `fixed-interval`? I think keeping these names is ok though if have a small explanation for each in documentation

alejoberardino: I think those are better, I'll rename and push a new commit here

amit-momin (approved):

PR #117, opened 2023-03-23
Add ability to override endpoint name in requester key

alejoberardino:


amit-momin (dismissed):

src/transports/http.ts
```diff
@@ -43,6 +43,9 @@ export type ProviderRequestConfig<T extends HttpTransportGenerics> = {
 
   /** The request that will be sent to the data provider to fetch values for the params in this struct */
   request: AxiosRequestConfig<T['Provider']['RequestBody']>
+
+  /** Whether to override the endpoint name for the requester key, to allow the coalescing of requests across endpoints */
```
amit-momin, line 47: Got to look this over again since this isn't urgent anymore and decided to nitpick. Think the comment is from when it was a flag
```suggestion
  /** An endpoint name override for the requester key to allow the coalescing of requests across endpoints */
```

amit-momin (approved):

PR #119, opened 2023-03-28
skip overwriting successful cache entries with errors

karen-stepanyan:
[PDI-2044](https://smartcontract-it.atlassian.net/browse/PDI-2044)

[PDI-2044]: https://smartcontract-it.atlassian.net/browse/PDI-2044?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Functional/manual tests were performed, both for local and redis to make sure that everything works as expected

### DIFF
--- a/src/cache/local.ts
+++ b/src/cache/local.ts
@@ -61,7 +61,7 @@ export class LocalCache<T = unknown> implements Cache<T> {
     }
   }
 
-  async set(key: string, value: T, ttl: number): Promise<void> {
+  async set(key: string, value: Readonly<T>, ttl: number): Promise<void> {
     logger.trace(`Setting key ${key} with ttl ${ttl}`)
     if (this.cache.has(key)) {
       logger.trace(`Found existing key ${key}. Updating value...`)
@@ -69,7 +69,7 @@ export class LocalCache<T = unknown> implements Cache<T> {
       const isCacheExpired = node.data.expirationTimestamp <= Date.now()
       // If the new value is an error, we don't want to overwrite the current active(not expired) 'successful' cache. If both cached and new values are errors or successful entries, the cache will be updated.
       if (
-        'errorMessage' in (value as AdapterResponse) &&
+        'errorMessage' in (value as unknown as AdapterResponse) &&
         !isCacheExpired &&
         !('errorMessage' in (node.data.value as AdapterResponse))
       ) {
--- a/src/cache/local.ts
+++ b/src/cache/local.ts
@@ -1,4 +1,4 @@
-import { DoubleLinkedList, LinkedListNode, makeLogger } from '../util'
+import { AdapterResponse, DoubleLinkedList, LinkedListNode, makeLogger } from '../util'
 import { Cache, CacheEntry } from './index'
 import { CacheTypes } from './metrics'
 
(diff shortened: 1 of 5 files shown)


src/cache/local.ts
```diff
@@ -61,17 +61,27 @@ export class LocalCache<T = unknown> implements Cache<T> {
     }
   }
 
-  async set(key: string, value: Readonly<T>, ttl: number): Promise<void> {
+  async set(key: string, value: T, ttl: number): Promise<void> {
```
alejoberardino, line 64: The `Readonly` here made sense, was it removed for a particular reason?

src/cache/redis.ts
```diff
@@ -15,7 +15,34 @@ const logger = makeLogger('RedisCache')
...
 
-  constructor(private client: Redis) {}
+  constructor(private client: Redis) {
+    this.loadFunctions()
+  }
+
+  async loadFunctions() {
+    // Load custom function 'setResponse' with lua script to redis that will skip overwriting successful cache response if the new value is error response
+    const source =
+      '#!lua name=ea \n ' +
```
alejoberardino, line 25: Let's change this to a multiline string (with `) so we don't have to sum every line

src/cache/redis.ts
```diff
@@ -15,7 +15,34 @@ const logger = makeLogger('RedisCache')
...
+      "redis.register_function('setResponse', function(keys, args) " +
```
alejoberardino, line 26: The name should be something more specific to decrease the risk of potential clashes (something like `ea_setResponse` or `setExternalAdapterResponse`)

src/cache/redis.ts
```diff
@@ -15,7 +15,34 @@ const logger = makeLogger('RedisCache')
...
+      'local key = keys[1] ' +
+      'local value = args[1] ' +
+      'local ttl = tonumber(args[2]) ' +
+      'local json_value = cjson.decode(value) ' +
+      "local key_exists = redis.call('EXISTS', key) " +
+      'if json_value.errorMessage and key_exists == 1 then ' +
```
alejoberardino, line 32: This is not checking if the existing value in the cache is an error or not, so it would not overwrite the error with a more recent one. Was this done because of a performance limitation, or is it doable?

src/cache/redis.ts
```diff
@@ -15,7 +15,34 @@ const logger = makeLogger('RedisCache')
...
+      "  return redis.call('SET', key, value, 'PX', ttl) " +
+      'end ' +
+      'end)'
+
+    const fnList = await this.client.function('LIST')
+    if (!fnList.length) {
+      this.client.function('LOAD', source)
+    } else {
+      this.client.function('LOAD', 'REPLACE', source)
+    }
```
alejoberardino, line 44: We cannot assume that we will be the only functions present, and the REPLACE option will override anyways, so the check is unnecessary. With the name change for the function, we can always do a function load replace

alejoberardino (changes requested):

src/cache/local.ts
karen-stepanyan: It was removed since i couldn't type cast `value as AdapterResponse` with `Readonly`. The other option was to convert it first to `unknown`

src/cache/redis.ts
karen-stepanyan: I changed it to also check for stored value as well. My concern was performance since if the new value is error we need to GET the current value. On the other hand it all happens inside a function and should be very fast. Also with this we have the same logic as for local cache

src/cache/local.ts
alejoberardino: I'd rather we cast to unknown in that case. We know what we're doing in here, but whoever uses the cache should be under strict types if possible

alejoberardino (changes requested):

alejoberardino (approved):

PR #129, opened 2023-04-04
censor sensitive values when writing to response cache

karen-stepanyan:
[PDI-180](https://smartcontract-it.atlassian.net/browse/PDI-180)

[PDI-180]: https://smartcontract-it.atlassian.net/browse/PDI-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/cache/response.ts
+++ b/src/cache/response.ts
@@ -1,10 +1,10 @@
 import { AdapterDependencies } from '../adapter'
 import { AdapterSettings } from '../config'
 import {
-  AdapterResponse,
+  AdapterResponse, censor,
   makeLogger,
   RequestGenerics,
-  ResponseGenerics,
+  ResponseGenerics, TimestampedAdapterResponse,
   TimestampedProviderErrorResponse,
   TimestampedProviderResult,
 } from '../util'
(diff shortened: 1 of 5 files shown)


src/cache/response.ts
```diff
@@ -56,9 +59,13 @@ export class ResponseCache<
    * @param results - the entries to write to the cache
    */
   async write(transportName: string, results: TimestampedProviderResult<T>[]): Promise<void> {
+    const censorList = CensorList.getAll()
     const entries = results.map((r) => {
+      const censoredResponse = censor(r.response, censorList) as TimestampedAdapterResponse<
```
alejoberardino, line 64: The censor function will not throw an error atm, and just respond with a redacted '[Unknown]'. This is not good enough for the EA responses, so we should split the censor fn into an internal one that we can use here that will float up the error so we can store a custom error response instead

alejoberardino (changes requested):

alejoberardino (approved):

PR #141, opened 2023-04-17
PDI 2010 - Added rate limit config validations for non-positive numbers in v3 framework

finleydecker:
Added a check in adapters/basic.ts for non-positive numbers in rate limit tiers. The framework now throws an error on startup if the rate limit value is not a positive number.

Also added a test in test/rate-limit-config.test.ts.


amit-momin (approved):
LGTM 🚀

PR #151, opened 2023-04-26
Refactor Input parameters

alejoberardino:
Input parameters in their current state do not support complex validations, and it’s complicated to enforce consistency between the transport/endpoint generics and the params themselves. This refactor changes them to a format similar to the AdapterConfig that allows for the same rough details, while making it easier to code and enforce types.


src/validation/input-params.ts
```diff
@@ -1,28 +1,475 @@
...
+type InputParameterOptionConstraints =
+  | {
+      // Only allow options for string params
+      type: 'string'
+      options?: string[] | readonly string[] // Enumerated options, ex. ['ADA', 'BTC', 'ETH']
+    }
+  | {
+      type: Exclude<ParameterType, 'string'>
+      options?: never
+    }
```
amit-momin, line 83: This was probably already the restriction but feel like there could be a use case for number options. Like if we wanted to restrict PoR adapter `chainId` params to specific ones for ethereum. Or if a data provider has a parameter to specify specific windows that are numbers.

alejoberardino: Fair enough, `chainId` is the counterargument to not allowing it. Changes pushed

src/validation/input-params.ts
```diff
@@ -1,28 +1,478 @@
...
+      options?: string[] | readonly string[]
+    }
+  | {
+      type: 'number'
+      options?: number[] | readonly number[]
+    }
+  | {
+      type: Exclude<ParameterType, 'string'>
+      options?: never
+    }
```
amit-momin, line 86: Wouldn't this need to be this now?
```suggestion
  | {
      type: Exclude<ParameterType, 'string' | 'number'>
      options?: never
    }
```

alejoberardino: Fixed

amit-momin (approved):

PR #156, opened 2023-04-27
Add smarter typing to optioned input params

alejoberardino:


src/validation/input-params.ts
```diff
@@ -37,6 +37,18 @@ type TypeFromTypeString<T extends ParameterType> = T extends 'string'
   ? TypeFromDefinition<T>
   : never
 
+/**
+ * All posible types for specifying options for an InputParameter
+ */
+type ParameterOptions = string[] | readonly string[] | number | readonly number[]
```
amit-momin, line 43: Should be this right?
```suggestion
type ParameterOptions = string[] | readonly string[] | number[] | readonly number[]
```

alejoberardino: Fixed

amit-momin (approved):

PR #164, opened 2023-04-27
PDI-2081 & 2082 -- Create scripts to parse EA settings and metrics and output markdown docs

finleydecker:
- Updated the Validator type to add a 'meta' object that exposes details about validators (later used in the EA-Settings table)
- Created two scripts to parse EA settings and metrics
- Each script outputs a markdown table stored in a new 'reference-tables' folder in the docs


src/validation/utils.ts
```diff
@@ -1,80 +1,129 @@
...
+  let validation = ''
+  
+  for (let i = 0; i < validatorFunctions.length; i++) {
+    const curValidator = validatorFunctions[i]
+    if (curValidator.meta.min !== undefined) { min = curValidator.meta.min }
+    if (curValidator.meta.max !== undefined) { max = curValidator.meta.max }
+    
+    validation += curValidator.meta.validation
+    if (i < validatorFunctions.length - 1) { validation += ', ' }
+  }
```
alejoberardino:
```suggestion
  const meta: ValidatorMeta = {}
  let details: string[] = []
  for (const validator of validators) {
    if (validator.meta.min) meta.min = validator.meta.min
    if (validator.meta.max) meta.max = validator.meta.max
    if (validator.meta.details) details.push(validator.meta.details)
  }
  meta.details = details.join(', ')
```

src/validation/utils.ts
```diff
@@ -1,80 +1,129 @@
 import { ValidationErrorMessage } from '../config'
 import { isIP } from 'net'
 
-export type Validator<V> = (value?: V) => ValidationErrorMessage
+export type Validator<V> = {
+  meta: {
+    min?: number
+    max?: number
+    validation?: string
```
alejoberardino, line 8: I think it's redundant to call this one validation, we're already within a Validator. I'd go back to sth like details or description

src/validation/utils.ts
```diff
@@ -89,23 +138,26 @@ const object = () => {
...
+    }
+    return compose(validators)
 }
 
 const port = () => integer({ min: 1, max: 65535 })
 
-// Validates that value is a valid timestamp from 2018-01-01 to now
-const responseTimestamp = () => integer({ min: 1514764861000, max: new Date().getTime() })
+// Validates that value is a valid timestamp from 2018-01-01 to now + 50ms to account for clock drift
+const responseTimestamp = () => integer({ min: 1514764861000, max: new Date().getTime() + 50 })
```
alejoberardino, line 158: Make sure to rebase your branch to main, these changes are already merged

src/validation/utils.ts
finleydecker: thanks I like this much better! 

Only thing I'm keeping is "curValidator.meta.min !== undefined"

If the min is 0 it won't be truthy, hence checking if it is undefined. I could change the if statement checking for the max since it is never 0, but I kind of like the verbosity in this case. what do you think?

alejoberardino (changes requested):

yarn.lock
```diff
@@ -1734,9 +1734,9 @@ map-age-cleaner@^0.1.3:
     p-defer "^1.0.0"
 
 marked@^4.0.19:
-  version "4.1.1"
-  resolved "https://registry.yarnpkg.com/marked/-/marked-4.1.1.tgz#2f709a4462abf65a283f2453dc1c42ab177d302e"
-  integrity sha512-0cNMnTcUJPxbA6uWmCmjWz4NJRe/0Xfk2NhXCUHjew9qJzFN20krFnsUe7QynwqOwa5m1fZ4UDg0ycKFVC0ccw==
+  version "4.3.0"
```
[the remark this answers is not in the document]
finleydecker: Hmm. I may have re-generated the lock file at some point. Should I not have done that? I still get the same lock-file when re-generating it now.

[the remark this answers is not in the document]
alejoberardino: You get a different lock file because dependencies of dependencies are fetched with the latest versions, but it shouldn't be part of a PR, I'm pushing it through so we can review faster

alejoberardino (approved):

yarn.lock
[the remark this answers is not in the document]
finleydecker: Gotcha. Thanks for the explanation

PR #167, opened 2023-04-29
Fix input validator 0 defaults

amit-momin:
- Fixed default value check not allowing 0 for number types
- Changed array validation logic to return the same value as passed in request. `undefined` doesn't return empty array.
  - Potentially up for discussion if we should change the behavior to this but I'd argue to maintain what it was before. Existing falsy checks on array request params would start failing (i.e. `if (req.addresses)`)

### DIFF
--- a/src/validation/input-params.ts
+++ b/src/validation/input-params.ts
@@ -328,7 +328,7 @@ class ProcessedParam<const T extends InputParameter = InputParameter> {
         if (this.definition.required || input != null) {
           throw this.validationError('input value must be a non-empty array')
         } else {
-          return input
+          return []
         }
       }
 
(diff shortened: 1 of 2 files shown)


alejoberardino:
The default 0 should be fixed, and it's true that it would be easier to return undefineds in arrays, but the logic in reality becomes more complex by doing so. Concretely, array variables:
- required
  - undefined or null: error (this one's easy)
  - empty array: so far there's been no validation for this, and it's confusing both in the logic and the types
  - array with at least one value: ok
- non-required
  - undefined or null: we want to accept this, but then in the logic we'd be dealing with either undefined, or a present array in the previous logic, with no way to tell.
  - empty array: accepted previously, in the current state too.

alejoberardino (approved):

PR #181, opened 2023-05-09
format ref tables

karen-stepanyan:
This PR modifies the ref-table generation script to also prettify it after generation.


amit-momin (approved):
I changed the label to none since this is just a script update but otherwise looks good!

PR #182, opened 2023-05-11
PDI-2131: Show request body on error

finleydecker:
- added a reqBody property to the errorWithContext object in the errorCatchingMiddleware function
Screenshot 2023-05-10 at 3 02 19 PM


amit-momin (approved):
Looking forward to this addition 🚀

PR #183, opened 2023-05-11
change the logic of version-bump pr workflow

karen-stepanyan:
[PDI-2099](https://smartcontract-it.atlassian.net/browse/PDI-2099)
This PR changes the logic for version-bump pr workflow. Previously each different labeled PR merges would create different Version bump PRs. For example given current version of 1.0.0, PR `A` labeled as `major` and PR `B` labeled as `patch` after the merge would create two version bump PRs, `version-bump/2.0.0-main` and `version-bump/1.0.1-main` respectively.

With the new logic a PR merge would create  just one version bump PR from  'version-bump' branch. In this case if other PRs are merged it will compare the versions and update the version bump PR if there is higher version change. 
The same example now. Given current version of 1.0.0, PR `A` is labeled as `major` and PR `B` labeled as `patch`. If `A` is merged first, version bump PR is created with version change from 1.0.0 to 2.0.0. After that `B` is merged, but since we already have a PR and version change from `B` (1.0.0 -> 1.0.1) is not higher (2.0.0) it will skip the step. 
If the `B` is merged first and then `A` then there will be two commits in the version bump PR, one changing version from 1.0.0 to 1.0.1 and second from 1.0.1 to 2.0.0.

The PR also modifies WORKFLOW-README.MD as there was outdated information about current actions and workflow

[PDI-2099]: https://smartcontract-it.atlassian.net/browse/PDI-2099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/open-version-bump-pr.yaml
+++ b/.github/workflows/open-version-bump-pr.yaml
@@ -80,7 +80,7 @@ jobs:
           echo "VERSION_TAG=$PACKAGE_VERSION" >> $GITHUB_OUTPUT
       - name: Create commit
         id: create-commit
-        if: inputs.dry-run != 'true'
+        if: inputs.dry-run != 'true' && ${{ steps.tick-version.outputs.VERSION_TAG != '' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           FILE_TO_COMMIT: package.json
@@ -174,6 +174,7 @@ jobs:
           fi
           echo "VERSION_BRANCH=$VERSION_BRANCH" >> $GITHUB_OUTPUT
       - name: Update code coverage badge
+        if: ${{ steps.tick-version.outputs.VERSION_TAG != '' }}
         continue-on-error: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -224,7 +225,7 @@ jobs:
           fi
       - name: Create PR
         id: create-pr
-        if: inputs.dry-run != 'true'
+        if: inputs.dry-run != 'true' && ${{ steps.tick-version.outputs.VERSION_TAG != '' }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
--- a/.github/workflows/open-version-bump-pr.yaml
+++ b/.github/workflows/open-version-bump-pr.yaml
@@ -84,41 +84,93 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           FILE_TO_COMMIT: package.json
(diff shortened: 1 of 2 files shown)


.github/workflows/open-version-bump-pr.yaml
```diff
@@ -84,41 +84,93 @@ jobs:
...
+              HIGHEST_VERSION="new"
+            else
+              # Compare patch version number
+              if (( bump[2] > new[2] )); then
+                HIGHEST_VERSION="bump"
+              elif (( bump[2] < new[2] )); then
+                HIGHEST_VERSION="new"
+              fi
+            fi
+          fi
```
amit-momin, line 136: If I'm understanding the logic correctly, couldn't we consolidate this down to one conditional? Something like `bump[0] < new[0] || bump[1] < new[1] || bump[2] < new[2]`

amit-momin: We could maybe just use that directly below where we check `HIGHEST_VERSION`

karen-stepanyan: I think this will not work for different levels. For example if the bump version is 1.1.5 and the new version is 1.0.9 it will assume that the new version is higher because of `|| bump[2] < new[2]` . Surely we can have all the conditions one-lined, but i left this for readability.
On the other hand i don't think we will have versions that will differ like this.

amit-momin: Oh ya you're right! This is good then

amit-momin:
I'm going to set the label to `none` since this isn't a change that needs to get deployed

.github/workflows/open-version-bump-pr.yaml
```diff
@@ -84,41 +84,93 @@ jobs:
...
+          
+          if [[ "$HIGHEST_VERSION" == "new" ]]; then
+           #new version is greater than current bump version, creating new commit 
+          
+           # move the branch pointer one commit backwards so that we can manually commit changes done by 'npm version ...' command
             git reset HEAD~
```
amit-momin, line 142: Would this still work now that we have the code-coverage badge commits?

karen-stepanyan: I hope so ) The code-coverage commits happen after version bump commit step so they are not really dependent on each other. I have tested that already.  Actually this PR has already created new [coverage-summary](https://github.com/smartcontractkit/ea-framework-js/blob/coverage-pr-183/coverage/coverage-summary.json) and the reason why i used `patch` label was just to run all the steps, have the PR created and see the new coverage result as well.

amit-momin: Ok sounds good! I'll add the label back then

amit-momin (approved):

amit-momin (approved):

PR #189, opened 2023-05-24
(workaround) Put logs that would expose the input parameters object behind a feature flag

finleydecker:
- Added an optional CENSOR_SENSITIVE_LOGS setting that defaults to false
- If set to true, sensitive logs will not be run

Linked ticket: https://smartcontract-it.atlassian.net/browse/PDI-2152

### DIFF
--- a/src/adapter/basic.ts
+++ b/src/adapter/basic.ts
@@ -215,6 +215,10 @@ export class Adapter<CustomSettingsDefinition extends SettingsDefinitionMap = Se
   initializeDependencies(inputDependencies?: Partial<AdapterDependencies>): AdapterDependencies {
     const dependencies = inputDependencies || {}
 
+    if (this.config.settings.LOG_INPUT_PARAMS === true) {
+      process.env['LOG_INPUT_PARAMS'] = 'true'
+    }
+
     if (
       this.config.settings.EA_MODE !== 'reader-writer' &&
       this.config.settings.CACHE_TYPE === 'local'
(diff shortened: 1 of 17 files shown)


alejoberardino (commented):
Have we checked that there's no other possible place where input parameters could be logged?

src/adapter/basic.ts
```diff
@@ -215,6 +221,10 @@ export class Adapter<CustomSettingsDefinition extends SettingsDefinitionMap = Se
   initializeDependencies(inputDependencies?: Partial<AdapterDependencies>): AdapterDependencies {
     const dependencies = inputDependencies || {}
 
+    if (this.config.settings.LOG_CENSOR === false) {
+      process.env['LOG_CENSOR'] = 'false'
+    }
```
alejoberardino, line 226: I don't get this part, the settings will be sourced from the env vars, why are we setting them back?

src/util/logger.ts
```diff
@@ -154,3 +154,10 @@ export function censor(obj: any, censorList: CensorKeyValue[], throwOnError = fa
   })
   return JSON.parse(result)
 }
+
+export const censorLogs = (logFunc: () => void) => {
+  if (process.env['LOG_CENSOR'] === 'false') {
```
alejoberardino, line 159: We should change the name of the setting to something more specific, e.g. CENSOR_SENSITIVE_LOGS

alejoberardino (changes requested):

PR #198, opened 2023-06-30
Scaffolding generator

karen-stepanyan:
[PDI-28](https://smartcontract-it.atlassian.net/browse/PDI-28)

## Scaffolding generator

This PR adds scaffolding generator to generate new v3 adapters.  [Yeoman](https://yeoman.io/) is used to create custom generator that can be called from monorepo. 
This is how Yeoman works. First we need to create custom **generator**, which is structured node package, or simply a folder with structured data. Once generator is ready it can be executed via `yo path_to_generator`. To have this functionality several changes were made.

### Changes

- There is now new `generator-adapter` folder in `scripts` directory. It contains all the generator logic and templates. 
- There is now new `adapter-generator.ts` file in `scripts` directory. It is the main entry point after monorepo runs `yarn generate-v3-adapter `. It will take all the arguments and call the `yo path_to_generator` to run the generator. 
- New node packages are added - `yeoman-generator` as dependency and `@types/yeoman-generator` as dev dep. 
- New command is added to `bin` in package.json. Essentially monorepo or other repos have to install `@chainlink/external-adapter-framework`  and `yo` packages and call `create-external-adapter [path: optional]` and that will call already mentioned adapter-generator.js which itself will run the generator. 
- New script is added to package.json called `build-generator`. Since generator logic and generator runner script are written in typescript the new command compiles it to javascript and moves them, alongside other necessary files into `dist/src`. In summary new `dist/src` will have following files 
    *current files and folders (adapter, config, metrics, ...)
    *adapter-generator.js
    *generator-adapter (this folder is identical to the one in the scripts folder, except main index.ts file is compiled to index.js)
- `build` script in package.json is modified to also build the generator. 
-  New step in added to main github workflow that builds example EA, builds and runs the tests
- Several docs are updated

### Generator features

After running the generator it will perform following tasks

1. Ask the name of adapter
2. Ask the number of endpoints
3. For each endpoint ask the endpoint name, endpoint aliases and transport types. User can select multiple transports
4. Ask if user wants helpful comments regarding different ea components to be included in the source code.
5. Ask for final confirmation 
6. Once confirmed based on provided input the generator will create  the folder, all the endpoints with their transports and integration tests 

https://github.com/smartcontractkit/ea-framework-js/assets/91897037/658f3802-ce80-48f0-8eca-1c3dd8a536b9

[PDI-28]: https://smartcontract-it.atlassian.net/browse/PDI-28?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/generator-adapter/generators/app/index.ts
+++ b/scripts/generator-adapter/generators/app/index.ts
@@ -31,6 +31,7 @@ module.exports = class extends Generator<{rootPath: string}> {
     // Whether to include explanation comments for different parts of adapter components
     includeComments: boolean
   }
+  endpointsAndAliases = new Set<string>()
   // When EXTERNAL_ADAPTER_GENERATOR_NO_INTERACTIVE is set to true, the generator will not prompt the user and will use the
   // default values to create one endpoint with all transports. This is useful for testing the generator in CI, or in cases
   // where the user wants to quickly generate the boilerplate code.
@@ -59,8 +60,12 @@ module.exports = class extends Generator<{rootPath: string}> {
     const endpoints: Record<string, GeneratorEndpointContext> = {}
 
     for (let i = 0; i < endpointCount; i++) {
-      let inputEndpointName = await this._promptEndpointName(i, endpoints)
-      let endpointAliases = await this._promptAliases(inputEndpointName, endpoints)
+      let inputEndpointName = await this._promptEndpointName(i)
+      this.endpointsAndAliases.add(inputEndpointName)
+
+      let endpointAliases = await this._promptAliases(inputEndpointName)
+      endpointAliases.forEach(alias => this.endpointsAndAliases.add(alias))
+
       const inputTransports = await this._promptTransports(inputEndpointName)
 
       endpoints[i] = {
@@ -323,7 +328,7 @@ module.exports = class extends Generator<{rootPath: string}> {
     return endpointCount
   }
 
-  private async _promptEndpointName(index, endpoints = {}): Promise<string> {
+  private async _promptEndpointName(index): Promise<string> {
     if (this.promptDisabled) {
       return 'price'
     }
@@ -338,19 +343,18 @@ module.exports = class extends Generator<{rootPath: string}> {
 
     if (endpointName === '') {
       this.log('Endpoint name cannot be empty')
-      return this._promptEndpointName(index, endpoints)
+      return this._promptEndpointName(index)
     }
 
-    const existingEndpoint = this._checkExistingEndpoint(endpoints, endpointName)
-    if (existingEndpoint) {
+    if (this.endpointsAndAliases.has(endpointName)) {
       this.log(`Endpoint named or aliased '${endpointName}' already exists`)
-      return this._promptEndpointName(index, endpoints)
+      return this._promptEndpointName(index)
     }
 
     return endpointName
   }
 
-  private async _promptAliases(inputEndpointName, endpoints = {}): Promise<string[]> {
+  private async _promptAliases(inputEndpointName): Promise<string[]> {
     if (this.promptDisabled) {
       return []
     }
@@ -360,19 +364,18 @@ module.exports = class extends Generator<{rootPath: string}> {
       message: `Comma separated aliases for endpoint '${inputEndpointName}':`,
       default: 'empty',
     })
-    let endpointAliases
+    let endpointAliases: string[]
 
     if (endpointAliasesAnswer === 'empty' || endpointAliasesAnswer.trim().length === 0) {
       return []
     } else {
-      endpointAliases = new Set(...[endpointAliasesAnswer.split(',').map(a => this._normalizeStringInput(a.trim()))])
-      endpointAliases = [...endpointAliases]
+      endpointAliases = [...new Set(...[endpointAliasesAnswer.split(',').map(a => this._normalizeStringInput(a.trim()))])]
     }
 
-    let existingEndpoint = this._checkExistingEndpointAliases(endpoints, endpointAliases)
+    let existingEndpoint = endpointAliases.some(a => this.endpointsAndAliases.has(a))
     if (existingEndpoint) {
-      this.log(`Endpoint '${existingEndpoint.inputEndpointName}' already contains one or more provided aliases.`)
-      return this._promptAliases(inputEndpointName, endpoints)
+      this.log(`One of endpoints already contains one or more provided aliases.`)
+      return this._promptAliases(inputEndpointName)
     }
     return endpointAliases
   }
@@ -443,26 +446,6 @@ module.exports = class extends Generator<{rootPath: string}> {
     }
 
     return useComments
-  }
-
-  // Based on user input for endpoint name check if there is already registered endpoint/aliases with that name
-  private _checkExistingEndpoint(endpoints: Record<string, GeneratorEndpointContext>, newEndpointNameInput: string): GeneratorEndpointContext | undefined {
-    return Object.values(endpoints).find(e => {
-      return e.inputEndpointName === newEndpointNameInput || e.endpointAliases.includes(newEndpointNameInput)
-    })
-  }
-
-  private _checkExistingEndpointAliases(endpoints: Record<string, GeneratorEndpointContext>, aliases: string[]): GeneratorEndpointContext | undefined {
-    let existingEndpoint
-    for (let i = 0; i < aliases.length; i++) {
-      const alias = aliases[i]
-      existingEndpoint = this._checkExistingEndpoint(endpoints, alias)
-      if (existingEndpoint) {
-        break
-      }
-    }
-    return existingEndpoint
-
   }
 
   //convert endpoint name to normalized name that can be used in imports/exports, i.e. crypto-one-two -> cryptoOneTwo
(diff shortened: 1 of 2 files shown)


scripts/generator-adapter/generators/app/index.ts
```diff
@@ -0,0 +1,298 @@
...
+        message: `Comma separated aliases for endpoint '${endpointName}':`,
+        default: "empty"
+      })
+      let endpointAliases
+
+      if (endpointAliasesAnswer === 'empty') {
+        endpointAliases = []
+      }else {
+        endpointAliases = endpointAliasesAnswer.split(',')
+      }
```
alejoberardino, line 85: Adds support for spaces just in case

```suggestion
      const endpointAliases = endpointAliasesAnswer === 'empty' ? [] : endpointAliasesAnswer.replace(/ /g, '').split(',')
```

scripts/generator-adapter/generators/app/index.ts
```diff
@@ -0,0 +1,298 @@
...
+            },
+            checked: true
+          },
+          {
+            name: "Websocket",
+            value: {
+              type: 'ws',
+              name: 'wsTransport'
+            }
+
```
alejoberardino, line 106: Could we add one here for a "custom" transport that outputs a static, testable NOP implementation? That is, a custom transport that implements all the necessary methods, and just returns a very basic hardcoded payload. Probably should be a foreground transport too

alejoberardino (changes requested):
Left some minor comments, a feature request, and I'll leave one more thing here: I think we should have a boolean choice in the beginning that asks "are you an expert in the EA framework? (i.e., do you want guide comments to be printed alongside the code?)" and if the users says they're not familiar, we print technically unnecessary but helpful comments. Things like explaining in the HTTP transport "hey, this prepareRequests method will take all the requests that have been sent into the adapter, and you should make the code return a list of requests that will be sent to the provider[...]". What do you think?

alejoberardino:
Also, this looks absolutely fantastic @karen-stepanyan. Awesome, awesome job.

karen-stepanyan:
Hey @alejoberardino . Thanks for your review and feedback. I think all the comments are addressed now, including two feature requests but i did some refactoring as well so i think this needs another review cycle. Also somewhat related to this ticket i updated the docs that describe endpoints, transports and tests.

scripts/generator-adapter/generators/app/index.ts
```diff
@@ -0,0 +1,441 @@
...
+          if (!existingEndpointRetry) {
+            break;
+          }
+          --retries
+
+          if (retries === 0) {
+            throw new Error(`There is already configured endpoint with name '${inputEndpointName}'`)
+          }
+        }
+      }
```
alejoberardino, line 62: I think chunks like this could be simplified, as well as ignoring max retries (just let the user quit eventually)
```typescript
endpointNames = new Set()

  ...

  
  promptEndpointName(index: number) {
    const { inputEndpointName } = await this.prompt<{inputEndpointName: string}>({
      type: "input",
      name: "inputEndpointName",
      message: `What is the name of endpoint #${index+1}:`,
      default: "price"
    })

    const endpointName = this._normalizeStringInput(inputEndpointName)

    if (endpointName === '') {
      console.log('Endpoint name cannot be empty')
      this.promptEndpointName(index)
    }
    
    if (this.endpointNames.has(endpointName)) {
      console.log(`Endpoint name ${endpointName} already exists`)
      this.promptEndpointName(index)
    }

    return endpointName
  }
```

alejoberardino: Then we can do
```suggestion
    for (let i = 0; i < endpointCount; i++) {
      const endpointName = await this._promptEndpointName(i)
```

alejoberardino: And same for the aliases chunk below

scripts/generator-adapter/generators/app/index.ts
```diff
@@ -0,0 +1,441 @@
...
+    }
+    return inputTransports
+  }
+
+  private async _promptConfirmation(adapterName: string, endpointNames: string): Promise<boolean> {
+    const { useComments }  = await this.prompt({
+      type: "confirm",
+      name: "useComments",
+      default: false,
+      message: `Do you want helpful comments to be included in the source code ?:`
```
alejoberardino, line 387:
```suggestion
      message: `Do you want helpful explicative comments to be included along with the source code? (These are usually not included with adapters but can be helpful if you're new to EA development):`
```

scripts/generator-adapter/generators/app/templates/src/endpoint/base.ts
```diff
@@ -0,0 +1,21 @@
+<% if (includeComments) { %>// Input parameters define the structure of the request expected by the endpoint.<% } %>
```
alejoberardino, line 1: I think these files could be stored with extensions like `*.ts.ejs`. This would allow VScode to do proper syntax highlighting, at least printing less errors

scripts/generator-adapter/generators/app/templates/src/endpoint/endpoint-router.ts
```diff
@@ -0,0 +1,33 @@
...
+import { SingleNumberResultResponse } from '@chainlink/external-adapter-framework/util'
+import { TransportRoutes } from '@chainlink/external-adapter-framework/transports'
+import { config } from '../config'
+import overrides from '../config/overrides.json'
+<% for(let i=0; i<inputTransports.length; i++) {%>
+import { <%= inputTransports[i].name %> } from '../transport/<%= inputEndpointName %>-<%= inputTransports[i].type %>' <% }
+%>
+
+<%- include ./base.ts %>
+<%/* [THIS WILL NOT BE RENDERED]. Indeed, below is a very unpleasant code, but the main purpose of it is to have everything nicely formatted, otherwise EJS puts unwanted new-lines for every 'if' placeholder */%>
```
alejoberardino, line 12: Not sure I follow 🤔

alejoberardino (changes requested):
Requested some more changes, and leaving something here that can either be done in this PR or a separate one: to ensure that these templates never break, I think we should have tests (separate from the main framework ones, mostly for CI) that generate adapters, check that TS builds (or just executes), and then run the integration tests that were generated as well. This would be a very very nice way to ensure that if any breaking changes are done to the framework, the templates are updated accordingly.

scripts/generator-adapter/generators/app/index.ts
karen-stepanyan: Hey @alejoberardino , is there any particular reason to use `Set` for endpoint+alias names ? Currently there are 
`_checkExistingEndpoint` and `_checkExistingEndpointAliases` that check if there is any endpoint or alias for given name. We could use Set just to make sure that aliases for the same endpoint don't contain duplicates but after that convert to array again since working with sets requires additional changes in the templates as well. 

so the same example you provided 
```typescript
if (this.endpointNames.has(endpointName)) {
      console.log(`Endpoint name ${endpointName} already exists`)
      this.promptEndpointName(index)
}
```
can be 
```typescript
const existingEndpoint = this._checkExistingEndpoint(endpoints, inputEndpointName)
    if (existingEndpoint) {
      this.log(`Endpoint named or aliased '${endpointName}' already exists`)
      return this._promptEndpointName(index, endpoints)
    }
```

scripts/generator-adapter/generators/app/templates/src/endpoint/base.ts
karen-stepanyan: Does it really work ? I'm using webstorm and changing to `.ts.ejs` actually makes it even worse. At least with `.ts` the colors and spacing for typescript code is correct, imports are navigating to correct paths. With  `.ts.ejs` for me everything looks like txt file.

scripts/generator-adapter/generators/app/templates/src/endpoint/endpoint-router.ts
karen-stepanyan: That is a comment that will NOT be included in the generated adapter. The syntax bellow that comment is really weird and could be simplified but the way ejs works is that it will create placeholder empty spaces and new lines for conditional statements . By default `if` statement creates new line character, so the rendered code would like something like this

```typescript
export const endpoint = new CryptoPriceEndpoint({
  name: 'crypto',

  aliases: ['crypto-batched', 'batched', 'batch', 'price'],

  transport,

  inputParameters: cryptoInputParams,

  overrides: overrides.coingecko,

})
```

There is a special workaround to render everything normally and ignore new lines. The comment there just tells to be careful when changing or formatting that part, as it will most likely break the final formatting.  This was the first place with this syntax, however other templates also have it, so i guess it applies for everything and comment can be removed.

scripts/generator-adapter/generators/app/templates/src/endpoint/base.ts
alejoberardino: When naming a template with `*.ts`, this is what I see:
image
I think regardless of whether there's support for ejs syntax, we should name the files with `.ejs` to better differentiate between template files and main src, that way we can easily filter and/or add rules for handling those within the IDE differently.

karen-stepanyan: Done

scripts/generator-adapter/generators/app/index.ts
alejoberardino: We can keep both structures, the set would be pretty minimal and make checks run in constant time (not a huge deal given the actual iterations, but would remove the loops from the code

PR #199, opened 2023-07-03
fix status_code for metrics

karen-stepanyan:
[PDI-226](https://smartcontract-it.atlassian.net/browse/PDI-2226)
This PR fixes bug with status_code property of http metrics where it  was 200 while response from EA was 4xx. Now the status_code will be the same as response status code if there are no errors.

### DIFF
--- a/src/validation/index.ts
+++ b/src/validation/index.ts
@@ -135,9 +135,18 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
 
 export const errorCatchingMiddleware = (err: Error, req: FastifyRequest, res: FastifyReply) => {
   // Add adapter or generic error to request meta for metrics use
-  // There's a chance we have no request context if there was an error during input validation
+  // There's a chance we have no request context if there was an error during input validation,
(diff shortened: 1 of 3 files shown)


src/validation/index.ts
```diff
@@ -135,9 +135,18 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
...
+  // There's a chance we have no request context if there was an error during input validation,
+  // but we still want to include error in meta for metrics
   if (req.requestContext) {
     req.requestContext.meta = { ...req.requestContext?.meta, error: err }
+  } else {
+    req.requestContext = {
+      cacheKey: '',
+      data: undefined,
+      endpointName: '',
+      transportName: '',
```
alejoberardino, line 147: Let's make sure this won't increase cardinality on labels, meaning if the default for cacheKey is '' or undefined, same for endpointName and transportName

alejoberardino (changes requested):

alejoberardino (approved):

PR #200, opened 2023-07-06
Allow logger injection

alejoberardino:
Once crucial piece of the EA framework is its logger, used consistently and often throughout the code. Due to the potential for applications that use the framework needing to conditionally modify or filter logs, this PR changes the current implementation to keep a static reference to a logger factory and defers the actual instancing until the first time a child logger is called.

Originally this was planned to be solved by implementing a Dependency Injection framework across all the code, but I've chosen to go with this approach to minify the impact to the current logic. From my research into the current alternatives (tsyringe, inversifyjs, typedi, etc) most of them use Decorators, which are still in an experimental stage and can be painful to work with once the codebase gets large enough.

### DIFF
--- a/src/adapter/basic.ts
+++ b/src/adapter/basic.ts
@@ -17,6 +17,7 @@ import { RateLimiterFactory, RateLimitingStrategy } from '../rate-limiting/facto
 import {
   AdapterRequest,
   AdapterResponse,
+  LoggerFactoryProvider,
   SubscriptionSetFactory,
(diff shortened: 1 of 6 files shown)


karen-stepanyan (approved):

finleydecker (approved):

KuphJr (approved):
LGTM! Should be pretty easy for me to implement once this gets released.

Thanks! 💙

PR #202, opened 2023-07-12
Make cacheKeyGenerator type specific to input params

alecgard:
Narrow the type for the cacheKeyGenerator function 'data' param to be specific to the input parameters. This allows cacheKeyGenerator implementations to safely know the structure of the data passed in.


alejoberardino (approved):
🚀

PR #207, opened 2023-07-13
PDI-2148: Allow multiple EA instances with different caches

finleydecker:
- removed two outdated scripts in package.json
- added the ability to configure a cache prefix
- added a test for the new cache prefixes
- implemented a cache locking mechanism using the node-redlock package
- added integration tests for the cache locking mechanism


src/cache/redis.ts
```diff
@@ -108,4 +109,37 @@ export class RedisCache<T = unknown> implements Cache<T> {
...
+      retryJitter: 200,
+    })
+
+    redlock.on('error', async (error) => {
+      if (error instanceof ResourceLockedError) {
+        logger.error(`Redlock error: ${error}`)
+        return
+      } else {
+        logger.error(`Redlock error: ${error}`)
+      }
```
alejoberardino, line 131:
```suggestion
    redlock.on('error', async (error) => {
      logger.error(`Redlock error: ${error}`)
```

test/cache/cache-key.test.ts
```diff
@@ -198,3 +198,77 @@ test.serial('throws error when cache data is not object', async (t) => {
...
+              },
+            }
+          }
+        })(),
+      }),
+      new AdapterEndpoint({
+        name: 'test-custom-cache-key',
+        cacheKeyGenerator: (_) => {
+          return `test:custom_cache_key`
+        },
```
alejoberardino, line 231: We should test one without the custom generator as well

test/cache/cache-lock.test.ts
```diff
@@ -0,0 +1,276 @@
...
+    const redisClient = new RedisMock() as unknown as Redis
+    const cache = new RedisCache(redisClient) // Fake redis
+    const dependencies: Partial<AdapterDependencies> = {
+      cache,
+      redisClient,
+    }
+
+    try {
+      await TestAdapter.start(adapter, t.context, dependencies)
+      await sleep(5000)
```
alejoberardino, line 264: We're not using mocked clocks here so this test will actually take 5s, let's change the lock duration to a much smaller number so the test takes 1-2s max

test/cache/cache-lock.test.ts
```diff
@@ -0,0 +1,276 @@
...
+      t.fail('An ExecutionError should have been thrown')
+    } catch (error) {
+      if (error instanceof ExecutionError) {
+        t.is(error.message, 'The operation was unable to achieve a quorum during its retry window.')
+      } else {
+        t.fail('An ExecutionError should have been thrown')
+      }
+    }
+  },
+)
```
alejoberardino, line 276: We need one more test where:
- an adapter starts
- acquires a lock
- immediately closes
- a second adapter with the same prefix starts, and fails to acquire the lock on the first try because the previous one is still valid
- after a few attempts, it eventually acquires the lock due to the first adapter failing to extend it

alejoberardino (changes requested):

src/adapter/basic.ts
```diff
@@ -114,6 +114,18 @@ export class Adapter<CustomSettingsDefinition extends SettingsDefinitionMap = Se
 
     this.dependencies = this.initializeDependencies(dependencies)
 
+    if (this.config.settings.EA_MODE !== 'reader' && this.dependencies.cache.lock) {
+      const redlockKey = this.config.settings.CACHE_PREFIX
```
alejoberardino, line 118: Let's remove any mention of redlock, keep it generic

src/cache/redis.ts
```diff
@@ -108,4 +110,53 @@ export class RedisCache<T = unknown> implements Cache<T> {
     // Record exec command sent to Redis
     recordRedisCommandMetric(CMD_SENT_STATUS.SUCCESS, 'exec')
   }
+
+  async lock(key: string, cacheLockDuration: number, retryCount: number): Promise<void> {
+    const start = Date.now()
+    const log = (msg: string) => `[${(Date.now() - start) / 1000}]: ${msg}`
+    const durationBuffer = cacheLockDuration * 0.7
```
alejoberardino, line 117: Because we want to cover the case where an adapter crashes without extending the lock, we need the total time spent retrying to necessarily be greater than the time the lock would be extended for.
Given the limitations cause by the redlock bug, I think we should handle retries ourselves which should be very simple, and would make it so there shouldn't be any issues when we acquire the first lock

src/cache/redis.ts
```diff
@@ -108,4 +110,53 @@ export class RedisCache<T = unknown> implements Cache<T> {
...
+      // Clear timeout on close for testing purposes
+      apiEmitter.on('onClose', () => {
+        clearTimeout(lockTimeout)
+      })
+    }
+
+    // Handing promise on purpose, infinite loop.
+    // We extend immediately after acquiring because redlock does not calculate
+    // the start time properly if it cannot acquire on the first attempt.
+    // TODO: add escape condition, otherwise tests won't work
```
alejoberardino, line 159: There is an event condition now, this comment is no longer necessary

src/index.ts
```diff
@@ -16,6 +17,7 @@ import { errorCatchingMiddleware, validatorMiddleware } from './validation'
 import { EmptyInputParameters } from './validation/input-params'
 
 export { FastifyInstance as ServerInstance }
+export const apiEmitter = new EventEmitter()
```
alejoberardino, line 20: I like the emitter approach, but let's create it in the adapter instead of making it a global var.
Additionally, let's rename it something like `shutdownNotifier` so that we clearly understand what events it's supposed to emit, imo apiEmitter is too generic

src/util/testing-utils.ts
```diff
@@ -139,6 +140,25 @@ export class RedisMock {
   setExternalAdapterResponse(key: string, value: string, ttl: number) {
     return this.set(key, value, 'PX', ttl)
   }
+
+  evalsha(...args: never[]) {
+    const redlockKey = args[2][0]
+    const instanceKey = args[2][1]
```
alejoberardino, line 146:
```suggestion
    const [redlockKey, instanceKey] = args[2]
```
_extremely_ minor suggestion

test/cache/cache-lock.test.ts
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    try {
+      await TestAdapter.start(adapter, t.context, dependencies)
+      await TestAdapter.start(adapter2, t.context, dependencies)
+
+      t.fail('An ExecutionError should have been thrown')
+    } catch (error) {
+      if (error instanceof ExecutionError) {
+        t.is(error.message, 'The operation was unable to achieve a quorum during its retry window.')
```
alejoberardino, line 73: If this is the error the adapter throws, we should catch it and rethrow it with something easier to understand ("the adapter failed to acquire a lock on the cache, check the cache prefix and make sure that if you're running two instances... etc. etc."

alejoberardino (changes requested):

src/util/testing-utils.ts
finleydecker: Because args is of type 'never[]' I would get the following error which would require more lines of code than just leaving as is...  how do we feel about using the type 'never'? ok in this scenario?

"Type never must have a [Symbol.iterator]() method that returns an iterator."

src/cache/redis.ts
```diff
@@ -111,52 +111,68 @@ export class RedisCache<T = unknown> implements Cache<T> {
...
+    key: string,
+    cacheLockDuration: number,
+    retryCount: number,
+    shutdownNotifier: EventEmitter,
+  ): Promise<void> {
     const start = Date.now()
     const log = (msg: string) => `[${(Date.now() - start) / 1000}]: ${msg}`
-    const durationBuffer = cacheLockDuration * 0.7
+    // Ensures lock is able to be extended in time
+    const durationBuffer = cacheLockDuration * 0.8
```
alejoberardino, line 123: Let's perform this calculation in the extension loop inline, we can explain it with a comment and it'll clearer there imo

alejoberardino (changes requested):
Very minor changes, almost ready to go

alejoberardino (approved):

PR #214, opened 2023-07-26
moved generating ref tables after version-tick

finleydecker:
createVersionBumpPR throwing error "npm ERR! Git working directory not clean."

The generate-ref-tables step is making changes, and the npm version command doesn't like that the git state is not clean. Moved the generate-ref-tables step to run after the version-tick.


alejoberardino (approved):
Let's try it

PR #215, opened 2023-07-28
Subclasses for PoR adapters

karen-stepanyan:
According to PoR composite adapter readme, we have 10 address/protocol adapters  and 16 balance/indexer adapters. 

This PR adds several helpful classes that can be used during the migration of PoR adapters to v3.  
Changes are 

1. new file `src/adapter/por.ts`. It includes new `PorAddressEndpoint` that address/protocol adapters should use. Functionality wise it's the same `AdapterEndpoint` with one difference. It ensures that the return type of EA contains some specific properties and is the same for all `PorAddressEndpoint`s. From my research it seems that all address/protocol adapters return similar data, some return even more information but at least all return properties that `PorAddressEndpoint` specifies as return type.  `PorAddressEndpoint` doesn't care about input params and their type as i couldn't find common things between input params of all address/protocol adapters. 
1.1. Most of address adapters use onchain calls and just 3 of them use http calls. Those 3 adapters can be used with HttpTransport, but others will need their own custom one. (similar to moonbeam which is already migrated)
2. PoRBalanceEndpoint and PoRTotalBalanceEndpoint
2.1.  new `PoRBalanceEndpoint` that should be used by most of balance/indexer adapters. Again it is the same  as `AdapterEndpoint`  but it requires specific input parameter types (which are also exported) and specific response type. Specifically `PoRBalanceEndpoint`s at least should have input param `addresses` which is an array of objects with `address` property. `PoRBalanceEndpoint`s  should return an array of objects as a result where each object at least contains `balance` property. 
2.2. new `PoRTotalBalanceEndpoint` that should be used by few adapters. There are 4 adapters currently that PoR supports as balance adapters but they don't return balance for an address,  they just return the total balance  for given addresses. `PoRTotalBalanceEndpoint` extends `PoRBalanceEndpoint` and changes the return type to be a numeric string. (but still requires same input params as `PoRBalanceEndpoint`).
3. new `PoRAdapter` class. It extends base `Adapter` and sets BACKGROUND_EXECUTE_TIMEOUT setting to the highest value. 

Summary 
Address/protocol adapters can safely use `PorAddressEndpoint`. Some of them can use HttpTransport, but for most custom transport is needed. 
Balance/indexer adapters can safely use `PoRBalanceEndpoint`. In terms of transports 
`blockchair` - can use HttpTransport (because it supports batching)
the rest, `ada-balance` , `avalanche`, `blockcypher`, `eth-balance`, `eth-beacon` , `lotus`,  `por-indexer`, `amberdata`, `blockchain.com`, `btc.com`, `cryptoapis`, `sochain` - need to use custom transport 
`polkadot-balance` and `stader-balance` - already migrated (custom transports)


amit-momin:
Well thought out and good research! I still need to look into it in more depth but I took a look at the adapters you mentioned in point 3. I wasn't able to find any live feeds still running them. I can get firmer confirmation but they may be unused in which case we don't have to support their setup in the framework.

src/adapter/address.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export type AddressEndpointGenerics = TransportGenerics & {
+  Response: PoRAddressResponse
+}
+
+/**
+ * An AddressEndpoint is a specific type of AdapterEndpoint. Meant to comply with standard practices for
+ * Data Feeds, its response type must be `PoRAddressResponse`
+ */
+export class AddressEndpoint<T extends AddressEndpointGenerics> extends AdapterEndpoint<T> {}
```
alejoberardino, line 28: We should add the prefix the prefix here too (`PoRAddressEndpoint`) so it's clearer

src/adapter/balance.ts
```diff
@@ -0,0 +1,87 @@
+import { InputParametersDefinition } from '../validation/input-params'
+import { TransportGenerics } from '../transports'
+import { AdapterEndpoint } from './endpoint'
+
+export type Balance = Record<string, unknown> & {
```
alejoberardino, line 5: Same as above, I think all these should have the PoR prefix as well

src/adapter/balance.ts
```diff
@@ -0,0 +1,87 @@
...
+export class BalanceEndpoint<T extends BalanceEndpointGenerics> extends AdapterEndpoint<T> {}
+
+/**
+ * Helper type structure that contains the different types passed to the generic parameters of TotalBalanceEndpoint
+ */
+export type TotalBalanceEndpointGenerics = BalanceEndpointGenerics & {
+  Response: {
+    Result: string
+    Data: {
+      result: string
```
alejoberardino, line 76: I think the result types might clash (string & Balance[]), it's ok if we keep these separate in that case

src/transports/balance.ts
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Transport implementation that is meant to be used with [[BalanceEndpoint]]. It takes incoming request that
+ * contains 'addresses' array as input data. Then each address object from that array is passed through `getBalance`
+ * function, that returns AxiosRequestConfig and `onResponse` function. The Data Provider response is passed through the
+ * `onResponse` function which should return [[Balance]] object. Once the balances of all addresses are fetched and processed,
+ * the transport will cache it and return as EA response.
+ * @typeParam T - all types related to the [[Transport]]
+ */
+export class BalanceTransport<T extends BalanceTransportGenerics> implements Transport<T> {
```
alejoberardino, line 161: I think I agree with your point about this not being super necessary, it might be easier to understand the adapters if they just implement a custom transport. To make them shorter they could extend some of the abstract classes like SubscriptionTransport if necessary

alejoberardino (changes requested):
Great job @karen-stepanyan, I left some comments but overall this is exactly the direction needed. @amit-momin post here what you find, but this PR won't need a lot more reviews after we know those findings imo

amit-momin:
Could we move forward with the PR without the BalanceTransport? Noticed it was removed in the latest commit anyways. I'm pretty sure those adapters aren't used looking through RDD but I'm still waiting on official confirmation. At least this way we can move forward with the migrations that don't need it and it can be added in separately if needed. What do you think? @alejoberardino

amit-momin:
Just got confirmation that those adapters aren't being used so we're good to proceed without the BalanceTransport

src/adapter/por.ts
```diff
@@ -0,0 +1,140 @@
...
+        e instanceof PoRAddressEndpoint,
+    )
+    if (!porEndpoints.length) {
+      throw new Error(`This PoRAdapter's list of endpoints does not contain a valid PoR endpoint`)
+    }
+
+    super(params)
+    // PoR requests take longer to process than normal feeds, that's why we set
+    // BACKGROUND_EXECUTE_TIMEOUT to the highest value
+    this.config.settings.BACKGROUND_EXECUTE_TIMEOUT = 180_000
```
alejoberardino, line 138: This would override any setting that the user sets; we should change the default instead before the EA is initialized (i.e. before calling super)

alejoberardino (changes requested):

alejoberardino (approved):

PR #221, opened 2023-08-03
Add settings debug endpoint

alejoberardino:
This PR introduces a the capability to enable debug endpoints for EAs. These aim to simplify the current triaging process, by allowing a quick query to the adapter itself to get the current settings. Any sensitive values are censored by the same process with which logs are censored.

The endpoints are optionally enabled with the `DEBUG_ENDPOINTS` env var.


finleydecker (changes requested):
Just one tiny thing in the tests, otherwise looks good to me!

test/debug-endpoints.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      required: true,
+      sensitive: true,
+      customSetting: true,
+      value: '[API_KEY REDACTED]',
+    },
+  )
+})
+
+test.serial('/debug/settings returns html response', async (t) => {
+  process.env['DEBUG_ENDPOINTS'] = 'true'
```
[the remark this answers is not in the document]
alejoberardino: Nice catch!

src/debug/settings-page.ts
```diff
@@ -0,0 +1,114 @@
...
+        </thead>
+        <tbody>
+          ${settings
+            .map(
+              (setting) => /* HTML */ `
+                <tr>
+                  <td>${setting.name}</td>
+                  <td>${setting.type}</td>
+                  <td>${setting.description}</td>
+                  <td class="${setting.default ? 'default' : ''}">${setting.value || ''}</td>
```
amit-momin, line 100: Is this styling to show if a certain value has a default?

test/debug-endpoints.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    {
+      type: 'string',
+      description: 'Api key',
+      name: 'API_KEY',
+      required: true,
+      sensitive: true,
+      customSetting: true,
+      value: '[API_KEY REDACTED]',
+    },
+  )
```
amit-momin, line 120: Just me being extra but could you add a couple more scenarios to test?
- Env var with default not explicitly set to show the true defaults showing up like `API_TIMEOUT` showing 30000
- Env var without default not explicitly set to show no value displayed like `RATE_LIMIT_API_TIER` showing no value

amit-momin (changes requested):

src/debug/settings-page.ts
alejoberardino: It basically reduces the opacity so it's easier to quickly see if an env var has a custom value or it's just using the default. It wasn't properly comparing tho, so I've updated it to check against the value as intended

src/debug/settings-page.ts
```diff
@@ -0,0 +1,114 @@
...
+      ...definitionDetails,
+      value: censor(value, censoredValues),
+    })
+  }
+  return censoredSettings
+}
+
+// To enable syntax highlighting in VSCode, you can download the "Comment tagged templates" extension:
+// https://marketplace.visualstudio.com/items?itemName=bierner.comment-tagged-templates
+const settingsPage = (settings: DebugPageSetting[]) => /* HTML */ `
```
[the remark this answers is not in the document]
alejoberardino: I don't think it's necessary, this is an extremely simple debug view and we don't really need a complex UI for it imo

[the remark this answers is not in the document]
amit-momin: I was thinking more so if they just had some standards or styling they wanted us to follow. But agree it might be overkill

test/debug-endpoints.test.ts
alejoberardino: Added tests

amit-momin (approved):

finleydecker (approved):

PR #225, opened 2023-08-15
Add metrics for request size, PoR addresses length

karen-stepanyan:
[PDI-2254](https://smartcontract-it.atlassian.net/browse/PDI-2254)

Changes

1. Added two new metrics, `requestPayloadSize` (histogram) and `porBalanceAddressLength` (counter)
2. Modified `PoRAdapter` to record length of addresses in metrics
3. Modified `buildMetricsMiddleware` middleware to record requestPayloadSize in metrics.

[PDI-2254]: https://smartcontract-it.atlassian.net/browse/PDI-2254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/adapter/por.ts
+++ b/src/adapter/por.ts
@@ -168,7 +168,7 @@ export class PoRAdapter<T extends SettingsDefinitionMap> extends Adapter<T> {
         metrics
           .get('porBalanceAddressLength')
           .labels({ feed_id: feedId })
-          .inc(data.addresses.length)
+          .set(data.addresses.length)
       }
     }
     return super.handleRequest(req, replySent)
--- a/src/metrics/index.ts
+++ b/src/metrics/index.ts
@@ -343,7 +343,7 @@ export const metrics = new Metrics(() => ({
     help: 'The number of data provider credits the adapter is consuming',
     labelNames: ['participant_id', 'feed_id'] as const,
   }),
-  porBalanceAddressLength: new client.Counter({
+  porBalanceAddressLength: new client.Gauge({
     name: 'por_balance_address_length',
     help: 'The number of addresses in PoR request input parameters',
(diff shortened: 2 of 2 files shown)


src/metrics/index.ts
```diff
@@ -335,4 +343,9 @@ export const metrics = new Metrics(() => ({
     help: 'The number of data provider credits the adapter is consuming',
     labelNames: ['participant_id', 'feed_id'] as const,
   }),
+  porBalanceAddressLength: new client.Counter({
```
amit-momin, line 346: I think we'd want a Gauge here. The addresses in a request could go up or down and we just want the amount in that particular request.

src/adapter/por.ts
```diff
@@ -152,4 +154,23 @@ export class PoRAdapter<T extends SettingsDefinitionMap> extends Adapter<T> {
...
+    const endpoint = this.endpoints.find((e) => e.name === req.requestContext.endpointName)
+
+    if (endpoint instanceof PoRBalanceEndpoint) {
+      const data = req.requestContext.data as { addresses: { address: string }[] }
+      if (data && data.addresses && Array.isArray(data.addresses)) {
+        const feedId = req.requestContext?.meta?.metrics?.feedId || 'N/A'
+        metrics
+          .get('porBalanceAddressLength')
+          .labels({ feed_id: feedId })
+          .inc(data.addresses.length)
```
amit-momin, line 171: Once this metric is moved to a Gauge, we'll need to update this to `set`
```suggestion
          .set(data.addresses.length)
```

amit-momin (changes requested):

PR #233, opened 2023-09-01
DF-19037 standard endpoint for LWBA

mmcallister-cll:
Some standardization of request/response format and aliases for LWBA endpoints.
Note that multiple LWBA endpoints are simply aliases of a crypto price endpoint.


austinborn (approved):
lgtm

austinborn (approved):

PR #237, opened 2023-10-05
DF-19232 set "Websocket url has changed" log to debug

mmcallister-cll:
The new type logs URL changes at debug level rather than info since frequent changes are expected.

https://smartcontract-it.atlassian.net/browse/DF-19232

### DIFF
--- a/src/transports/websocket.ts
+++ b/src/transports/websocket.ts
@@ -129,7 +129,6 @@ export class WebSocketTransport<
   currentUrl = ''
   lastMessageReceivedAt = 0
   connectionOpenedAt = 0
-  isUrlSubscriptionWebsocket = false
 
   constructor(private config: WebSocketTransportConfig<T>) {
     super()
--- a/src/transports/websocket.ts
+++ b/src/transports/websocket.ts
@@ -327,11 +327,19 @@ export class WebSocketTransport<
 
     // Check if we should close the current connection
     if (!connectionClosed && (urlChanged || connectionUnresponsive)) {
-      const reason = urlChanged
-        ? `Websocket url has changed from ${this.currentUrl} to ${urlFromConfig}, closing connection...`
-        : `Last message was received ${timeSinceLastMessage} ago, exceeding the threshold of ${context.adapterSettings.WS_SUBSCRIPTION_UNRESPONSIVE_TTL}ms, closing connection...`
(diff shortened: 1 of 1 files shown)


src/transports/websocket.ts
```diff
@@ -329,7 +330,8 @@ export class WebSocketTransport<
       const reason = urlChanged
         ? `Websocket url has changed from ${this.currentUrl} to ${urlFromConfig}, closing connection...`
         : `Last message was received ${timeSinceLastMessage} ago, exceeding the threshold of ${context.adapterSettings.WS_SUBSCRIPTION_UNRESPONSIVE_TTL}ms, closing connection...`
-      censorLogs(() => logger.info(reason))
+      const logfn = urlChanged && this.isUrlSubscriptionWebsocket ? logger.debug : logger.info
+      censorLogs(() => logfn(reason))
```
alecgard, line 334: It looks like the `Last message was received...` message would be useful to receive as an `info` here regardless of if we're using URL subscription or not. 

Would there be any issue with simply making `Websocket url has changed...` always `debug` and `Last message was received...` always `info` and not having a separate class?

mmcallister-cll: Yeah, it's not a problem to make it always `debug` level (would also only make this a `patch` version update), but I thought it may be useful to maintain `info` for other EAs that don't require constant URL switching on subscription change. Maybe that's not useful though 🤷 

The only log that can be set to `debug` here is when the new class is explicitly used AND the url is changed (ie: the `Websocket url has changed...`).

alecgard: Personally I think the `Websocket url has changed...` fits `debug`, and it'd be nice to not have the complexity of multiple classes. 

Happy to go with the current approach if you feel strongly though!

mmcallister-cll: sounds good, happy to change it as well.

src/transports/websocket.ts
```diff
@@ -129,6 +129,7 @@ export class WebSocketTransport<
   currentUrl = ''
   lastMessageReceivedAt = 0
   connectionOpenedAt = 0
+  isUrlSubscriptionWebsocket = false
```
alecgard, line 132:
```suggestion
```

alecgard (approved):

PR #247, opened 2024-03-19
Added PoRProvider types

karen-stepanyan:
[DF-19638](https://smartcontract-it.atlassian.net/browse/DF-19638)
Added new types for new `PoRProviderEndpoint`. `PoRProviderEndpoint` enforces the response of the endpoint to have `result` and `ripcord` properties

[DF-19638]: https://smartcontract-it.atlassian.net/browse/DF-19638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/adapter/por.ts
+++ b/src/adapter/por.ts
@@ -134,7 +134,9 @@ export type PoRProviderEndpointGenerics = TransportGenerics & {
  * A PoRProviderEndpoint is a specific type of AdapterEndpoint. Meant to comply with standard practices for
  * Data Feeds, the response type must be `PoRProviderResponse`.
  */
-export class PoRProviderEndpoint<T extends PoRProviderEndpointGenerics> extends AdapterEndpoint<T> {}
+export class PoRProviderEndpoint<
+  T extends PoRProviderEndpointGenerics,
+> extends AdapterEndpoint<T> {}
 
 /**
  * A PoRAdapter is a specific kind of Adapter that includes at least one PoRTotalBalanceEndpoint,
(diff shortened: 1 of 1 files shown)


austinborn (approved):

src/adapter/por.ts
```diff
@@ -7,6 +7,14 @@ import { EmptyInputParameters, InputParametersDefinition } from '../validation/i
 import { AdapterRequest, AdapterResponse } from '../util'
 import { metrics } from '../metrics'
 
+export type PoRProviderResponse = {
+  Result: number
+  Data: {
+    result: number
+    ripcord: boolean
```
alecgard, line 14: Is there a `ripcordMessage` that we can rely on always being included that provides details on why the ripcord is true?

Not a blocker, as we probably don't want to enforce that as a required property (but still include when available)

alecgard (approved):

src/adapter/por.ts
karen-stepanyan: In theory there should be `ripcordDetails` but i wasn't able to get a response with `ripcord=true` to check whether details are provided or not. I think once we have multi-variable feeds, we can logger.warn the ripcordDetails. Not sure if it makes sense to actually return that value as response. In any case the types here just require `result` and `ripcord`, additional ones can be added.

alecgard: That sounds good to me 👍

PR #260, opened 2024-07-08
Introduce exponential backoff for subscription errors

austinborn:
- Add configurable `STREAM_HANDLER_RETRY_MIN_MS`, `STREAM_HANDLER_RETRY_MAX_MS` and `STREAM_HANDLER_RETRY_EXP_FACTOR` for exponential retries for subscription handlers. Default to `100ms * 3^retryCount` (100ms, 300ms, ~1s, 3s, 8s, 24s, ..., 20m) 
- Remove commented config vars that don't seem to be used at the moment

### DIFF
--- a/.github/workflows/pinned-dependencies.yaml
+++ b/.github/workflows/pinned-dependencies.yaml
@@ -10,7 +10,7 @@ jobs:
         uses: sergeysova/jq-action@a3f0d4ff59cc1dddf023fc0b325dd75b10deec58 # v2.3.0
         id: versions
         with:
-          cmd: "jq '[getpath([\"dependencies\"],[\"devDependencies\"],[\"optionalDependencies\"])] | del(..|nulls) | map(.[]) | join(\",\")' package.json -r"
+          cmd: 'jq ''[getpath(["dependencies"],["devDependencies"],["optionalDependencies"])] | del(..|nulls) | map(.[]) | join(",")'' package.json -r'
           multiline: true
       - name: Check for un-pinned versions
         run: |
--- a/.github/workflows/pinned-dependencies.yaml
+++ b/.github/workflows/pinned-dependencies.yaml
@@ -10,7 +10,7 @@ jobs:
         uses: sergeysova/jq-action@a3f0d4ff59cc1dddf023fc0b325dd75b10deec58 # v2.3.0
         id: versions
         with:
-          cmd: 'jq ''[getpath(["dependencies"],["devDependencies"],["optionalDependencies"])] | del(..|nulls) | map(.[]) | join(",")'' package.json -r'
+          cmd: "jq '[getpath([\"dependencies\"],[\"devDependencies\"],[\"optionalDependencies\"])] | del(..|nulls) | map(.[]) | join(\",\")' package.json -r"
           multiline: true
       - name: Check for un-pinned versions
         run: |
--- a/src/background-executor.ts
+++ b/src/background-executor.ts
@@ -92,10 +92,10 @@ export async function callBackgroundExecutes(adapter: Adapter, apiShutdownPromis
       // This background execute loop is no longer the one to determine the sleep between bg execute calls.
       // That is now instead responsibility of each transport, to allow for custom ones to implement their own timings.
       logger.trace(
-        `Finished background execute for endpoint "${endpoint.name}", calling it again in 10ms...`,
+        `Finished background execute for endpoint "${endpoint.name}", calling it again in ${adapter.config.settings.BACKGROUND_EXECUTE_BUFFER_INTERVAL_MS}ms...`,
       )
       metricsTimer()
-      timeoutsMap[endpoint.name] = setTimeout(handler, 10)
+      timeoutsMap[endpoint.name] = setTimeout(handler, adapter.config.settings.BACKGROUND_EXECUTE_BUFFER_INTERVAL_MS)
     }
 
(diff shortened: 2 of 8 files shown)


Subarna-Singh (dismissed):

austinborn:
Noticed the lint and test checks fail, will debug those before moving forward with this

src/background-executor.ts
```diff
@@ -92,10 +92,10 @@ export async function callBackgroundExecutes(adapter: Adapter, apiShutdownPromis
       // This background execute loop is no longer the one to determine the sleep between bg execute calls.
       // That is now instead responsibility of each transport, to allow for custom ones to implement their own timings.
       logger.trace(
-        `Finished background execute for endpoint "${endpoint.name}", calling it again in 10ms...`,
+        `Finished background execute for endpoint "${endpoint.name}", calling it again in ${adapter.config.settings.BACKGROUND_EXECUTE_BUFFER_INTERVAL_MS}ms...`,
       )
       metricsTimer()
-      timeoutsMap[endpoint.name] = setTimeout(handler, 10)
+      timeoutsMap[endpoint.name] = setTimeout(handler, adapter.config.settings.BACKGROUND_EXECUTE_BUFFER_INTERVAL_MS)
```
alejoberardino, line 98: We removed this capability intentionally to delegate this timing to the transports themselves + remove the extra risk of misconfiguration from an additional variable, I'd recommend against adding this in

src/transports/abstract/subscription.ts
```diff
@@ -68,7 +78,16 @@ export abstract class SubscriptionTransport<const T extends TransportGenerics>
       })
       .set(entries.length)
 
-    await this.backgroundHandler(context, entries)
+    try {
+      await this.backgroundHandler(context, entries)
+      this.retryCount = 0
```
alejoberardino, line 83: This is tracking state in the transport itself and there is a non-zero chance that a call to the bg handler takes more than the timeout and the framework will retry before that promise resolves; not critical here in practice but worth mentioning

src/transports/abstract/subscription.ts
```diff
@@ -54,6 +59,11 @@ export abstract class SubscriptionTransport<const T extends TransportGenerics>
 
   async backgroundExecute(context: EndpointContext<T>): Promise<void> {
     logger.debug('Starting background execute')
+
+    if (this.retryTime > Date.now()) {
+      return
+    }
+
     const entries = await this.subscriptionSet.getAll()
```
alejoberardino, line 67: I'm curious about the motivations behind this change, any additional env vars are new risks and I'm wondering if this can be solved differently

alejoberardino (commented):
Not a stakeholder in this so not relevant for me to request changes, just leaving comments with my thoughts

src/background-executor.ts
austinborn: Good point, I can remove this variable for now (if it's not broken, don't fix it)

src/transports/abstract/subscription.ts
austinborn: I see the point, although I agree I don't think this is an issue here. The retry only matters if the handler will error out, but I don't think we want to change `retryCount` and `retryTime` until we _know_ that a `backgroundHandler` call has failed. The biggest risk is if network requests take a while to respond in the `backgroundHandler` and _then_ it fails, but at that point it will increment the `retryCount` and start limiting additional requests.

src/transports/abstract/subscription.ts
austinborn: An alternative to this might be to set a specific retry limit per data provider, although exponential back-off was specifically requested and should help alleviate issues across all DPs.

.github/workflows/pinned-dependencies.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
         uses: sergeysova/jq-action@a3f0d4ff59cc1dddf023fc0b325dd75b10deec58 # v2.3.0
         id: versions
         with:
-          cmd: "jq '[getpath([\"dependencies\"],[\"devDependencies\"],[\"optionalDependencies\"])] | del(..|nulls) | map(.[]) | join(\",\")' package.json -r"
```
mmcallister-cll, line 13: are we good with this or should we remove?

src/config/index.ts
```diff
@@ -112,6 +102,26 @@ export const BaseSettingsDefinition = {
     type: 'string',
     default: '',
   },
+  SUBSCRIPTION_RETRY_MAX_MS: {
```
mmcallister-cll, line 105: Is this the right name for this env var?
Asking because WS DPs usually require a _connection_ step and then a _subscribe_ (to assets) step. I'm worried this might mix up background execute subscriptions with the actual subscribe action after a WS connection.

.github/workflows/pinned-dependencies.yaml
austinborn: `yarn prettier` was complaining about it, not sure if this was a recent change for `prettier` or something else...

austinborn: I'll try reverting this and see if it's ok

src/config/index.ts
austinborn: Updated to `STREAM_HANDLER_RETRY_` since we've moved it into the stream transport

mmcallister-cll (approved):

PR #333, opened 2024-11-29
Bump version to 1.7.2

mxiao-cll:
Mostly dependency version bumps


mxiao-cll:
The test is expected to fail because generated adapter is looking for 1.7.2 which doesn't exist yet

mmcallister-cll (approved):

PR #339, opened 2024-12-03
Fix logging

mxiao-cll:
https://stackoverflow.com/questions/71938587/unable-to-determine-transport-target-for-pino-pretty

For some reason it is now throwing error in deployments after version bump


mmcallister-cll (approved):

src/util/logger.ts
```diff
@@ -13,48 +14,47 @@ export type Store = {
...
         }
       }
-    }
-    return {}
+      return {}
+    },
   },
-  transport: process.env['DEBUG'] === 'true' ? debugTransport : undefined,
-})
+  process.env['DEBUG'] === 'true' ? stream : undefined,
```
mmcallister-cll, line 56: you don't need to set transport here?

PR #368, opened 2025-01-13
Disable includes.json files affects for endpoint:crypto/ crypto_lwba aliases

Subarna-Singh:
DF 20953 - RWA updates- Disable includes.json files affects for endpoint:crypto/ crypto_lwba aliases
https://smartcontract-it.atlassian.net/browse/DF-20953

### DIFF
--- a/src/adapter/price.ts
+++ b/src/adapter/price.ts
@@ -141,12 +141,6 @@ export class PriceAdapter<
         }
         const includesDetails = this.includesMap?.[requestData.base]?.[requestData.quote]
 
-        if (includesDetails?.endpoints?.length === 0) {
-          throw new Error(
-            `No endpoints supported in includes.json for ${requestData.base}/${requestData.quote}.`,
-          )
-        }
-
         if (
           includesDetails?.endpoints === undefined ||
           includesDetails?.endpoints?.includes(req.requestContext.endpointName)
--- a/src/adapter/price.ts
+++ b/src/adapter/price.ts
@@ -141,12 +141,16 @@ export class PriceAdapter<
         }
         const includesDetails = this.includesMap?.[requestData.base]?.[requestData.quote]
 
-        if (includesDetails?.endpoints.length === 0){
-          throw new Error(`No endpoints supported in includes.json for ${requestData.base}/${requestData.quote}.`)
+        if (includesDetails?.endpoints.length === 0) {
+          throw new Error(
+            `No endpoints supported in includes.json for ${requestData.base}/${requestData.quote}.`,
+          )
         }
 
-        if (!includesDetails?.endpoints.includes(req.requestContext.endpointName)){
-          throw new Error(`Endpoint ${req.requestContext.endpointName} not supported for ${requestData.base}/${requestData.quote} in includes.json`)
+        if (!includesDetails?.endpoints.includes(req.requestContext.endpointName)) {
+          throw new Error(
+            `Endpoint ${req.requestContext.endpointName} not supported for ${requestData.base}/${requestData.quote} in includes.json`,
+          )
         }
 
         if (includesDetails) {
@@ -161,7 +165,7 @@ export class PriceAdapter<
       }
 
       for (const endpoint of priceEndpoints) {
-          endpoint.requestTransforms?.push(requestTransform)
+        endpoint.requestTransforms?.push(requestTransform)
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/adapter/price.ts
```diff
@@ -152,7 +152,9 @@ export class PriceAdapter<
       }
 
       for (const endpoint of priceEndpoints) {
-        endpoint.requestTransforms?.push(requestTransform)
+        if (!['crypto', 'crypto-lwba'].includes(endpoint.name)){
```
mxiao-cll, line 155: Recommend we lowercase it to avoid issues with for example Crypto

And is this safe? I see that for cryptocompare we are overrding BTC and ETH

alecgard (dismissed):

src/adapter/price.ts
mmcallister-cll: agree here, i think hardcoding `'crypto', 'crypto-lwba'` might also miss endpoint name aliases in EAs where the "standard" endpoint name isn't used (eg: `price` instead of `crypto`)

For a more complete solution, my initial thought is something like eg:
```json
[
  {
    "from": "BTC",
    "to": "ETH",
    "includes": [
      {
        "from": "ETH",
        "to": "BTC",
        "inverse": true,
        "endpoints": ["crypto", "crypto-lwba"]
      }
    ]
  }
]
```

where `endpoints` defaults to all endpoints - current behaviour - for backward compatibility

Subarna-Singh: @mmcallister-cll I agree with the implementation idea on including endpoints in includes.json. However, I am not in favour of default to all endpoints because that would defeat the purpose of this proposed change. 

We want to enforce inclused.json to forex/RWA assets only. https://smartcontract-it.atlassian.net/browse/DF-20953

incase an EA wants to use includes.json for crypto endpoint it should exclusively specify so. 

I'll verify which EAs are impacted by this so we can make an informed decision. If the impact is small we can enforce the change without needing backward compatibility.

src/adapter/price.ts
```diff
@@ -140,14 +141,29 @@ export class PriceAdapter<
         }
         const includesDetails = this.includesMap?.[requestData.base]?.[requestData.quote]
 
-        if (includesDetails) {
-          requestData.base = includesDetails.from || requestData.base
-          requestData.quote = includesDetails.to || requestData.quote
+        if (includesDetails?.endpoints?.length === 0) {
+          throw new Error(
+            `No endpoints supported in includes.json for ${requestData.base}/${requestData.quote}.`,
+          )
```
mxiao-cll, line 147: If length is 0 then includesDetails?.endpoints?.includes(req.requestContext.endpointName) will always be false right? So we don't really need this block of logic?

mxiao-cll: In that case 
```
includesDetails?.endpoints?.includes(req.requestContext.endpointName)
```
will be false then we will throw error

Subarna-Singh: We do that here.
https://github.com/smartcontractkit/ea-framework-js/blob/e5a897b86d2925e94ffeb3f17e778a804c94ea2a/src/adapter/price.ts#L163

If endpoints is undefined we want the default behaviour to be support all endpoints to support backward compatibility. Hence, the structure of the code.

Cases covered
Undefined endpoints: support all endpoints
Endpoints field included but empty: throw error
Endpoints field included and endpoint included in it: process request
Endpoints field included and endpoint not included in it: throw error

mxiao-cll: Right my point is that 163 already covers 144-147 so you don't need 144 - 147

Subarna-Singh: Check.

mxiao-cll (approved):
I think logic is good, can we add some tests maybe?

src/adapter/price.ts
```diff
@@ -140,14 +141,23 @@ export class PriceAdapter<
...
-          requestData.base = includesDetails.from || requestData.base
-          requestData.quote = includesDetails.to || requestData.quote
-        }
-
-        const inverse = includesDetails?.inverse || false
-        priceRequest.requestContext.priceMeta = {
-          inverse,
+        if (
+          includesDetails?.endpoints === undefined ||
+          includesDetails?.endpoints?.includes(req.requestContext.endpointName)
```
mmcallister-cll, line 146: does this line not cover 145 with the `endpoints?`?

mxiao-cll (approved):

PR #382, opened 2025-01-27
Bump websocket connect timeout

akuzni2:
We need to be able to adjust the timeout level to be longer if some EA's are under heavy load we can still connect without doing multiple re-attempts

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -246,7 +246,7 @@ export const BaseSettingsDefinition = {
       'The maximum amount of time in milliseconds to wait for the websocket connection to open (including custom open handler)',
     type: 'number',
     default: 10_000,
-    validate: validator.integer({ min: 500, max: 30_000 }),
+    validate: validator.integer({ min: 500, max: 180000 }),
   },
   CACHE_POLLING_MAX_RETRIES: {
     description:
--- a/src/transports/websocket.ts
+++ b/src/transports/websocket.ts
@@ -157,7 +157,7 @@ export class WebSocketTransport<
     return {
       // Called when the WS connection is opened
       open: async (event: WebSocket.Event) => {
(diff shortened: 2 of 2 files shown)


src/transports/websocket.ts
```diff
@@ -157,7 +157,7 @@ export class WebSocketTransport<
     return {
       // Called when the WS connection is opened
       open: async (event: WebSocket.Event) => {
-        logger.debug(`Opened websocket connection. (event type ${event.type})`)
+        logger.info(`Opened websocket connection. (event type ${event.type})`)
```
akuzni2, line 160: This is an important type of event that is good to set at info level since it should be infrequent and manages the entire stream of data

src/config/index.ts
```diff
@@ -246,7 +246,7 @@ export const BaseSettingsDefinition = {
       'The maximum amount of time in milliseconds to wait for the websocket connection to open (including custom open handler)',
     type: 'number',
     default: 10_000,
-    validate: validator.integer({ min: 500, max: 30_000 }),
+    validate: validator.integer({ min: 500, max: 180000 }),
```
mmcallister-cll, line 249: not opposed to this per se, but also not sure this is correct, discussion: https://chainlink-core.slack.com/archives/C02FM5ZP2TC/p1738019106947369

src/transports/websocket.ts
mmcallister-cll: there are certain EAs (eg: coinmetrics) that require reconnection to change the subscription set. I'm not sure this is the correct approach. Maybe in the specific EA if you find this valuable to log at `info`, stick a separate log in the specific `open` handler?

akuzni2: That's a good suggestion - will implement locally

PR #383, opened 2025-01-29
remove duplicated metric capture

akuzni2:
Cache metrics are being set twice here when using redis cache type:

Response cache will call `setMany`[here](https://github.com/smartcontractkit/ea-framework-js/blob/6b83e6d61932a5c8b4f6d92e969c584e6942866d/src/cache/response.ts#L135)

In the Redis implementation that tracks various metrics [here](https://github.com/smartcontractkit/ea-framework-js/blob/6b83e6d61932a5c8b4f6d92e969c584e6942866d/src/cache/redis.ts#L112-L125)

Once `setMany` returns - we track the metrics again [here](https://github.com/smartcontractkit/ea-framework-js/blob/6b83e6d61932a5c8b4f6d92e969c584e6942866d/src/cache/response.ts#L137-L150)

### DIFF
--- a/src/cache/redis.ts
+++ b/src/cache/redis.ts
@@ -108,22 +108,6 @@ export class RedisCache<T = unknown> implements Cache<T> {
 
     await chain.exec()
 
-    // Loop again, but this time to record these in metrics
-    const now = Date.now()
-    for (const entry of entries) {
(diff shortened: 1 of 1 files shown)


mxiao-cll:
Fix tests

mxiao-cll (approved):
Tested locally?

PR #442, opened 2025-03-21
Allow hyphen in override adapter name

dskloetc:
[DF-21153](https://smartcontract-it.atlassian.net/browse/DF-21153)

# Description

The name of an adapter can refer to 2 different things:
1. The name of the package that implements the adapter
2. The name field in the definition of the adapter.

These are typically (but not necessarily) related but there are difference:
1. The package name is typically lower case while the name field is required to be upper case.
2. The package name typically uses a hyphen as separator while the name field typically uses an underscore as separator.

See for example the [name field](https://github.com/smartcontractkit/external-adapters-js/blob/cb6a8a287e311fd17697810c0accd2e2a093e43b/packages/sources/blocksize-capital/src/index.ts#L7) of the `blocksize-capital` adapter which has value `BLOCKSIZE_CAPITAL`.

This causes confusion when specifying request overrides for an adapter because the adapter name key used has to match the name field of the adapter but in lower case. Usually this is the same as the package name, except when that would have a hyphen, it has to be an underscore.

To avoid confusion, we want to allow both hyphen and underscore to be used, to make it more likely that the package name will work as override key.

# Changes

1. Add `getCanonicalAdapterName` which will convert an adapter name to lower case and replace hyphens with underscores.
2. Add `canonicalizeAdapterNameKeys` which will take a record with adapter names as keys and canonicalize all the keys.
3. In `getRequestOverrides` canonicalize both the lookup key and the keys in the request overrides.
4. In `defaultRouter` reuse `getRequestOverrides` so the canonicalizing logic doesn't need to be duplicated. Note that when  `defaultRouter` is called, `adapterNameOverride` is not available yet, so it can't reuse the full logic of `getRequestOverrides`.

# Testing

1. In `test/overrides.test.ts` and `test/transports/routing.test.ts` change the test adapter name from `TEST` to `TEST_ADAPTER` to allow testing with hyphens.
2. Add tests for overriding base assets and transport with a hyphen.
3. Add unit tests for the utility functions.
4. Tested manually in `external-adapters-js` by using a [portal dependency](https://github.com/smartcontractkit/external-adapters-js/blob/main/CONTRIBUTING.md#framework-development) and using the following request while running the `blocksize-capital` EA:
```
$ curl -X POST http://localhost:8080 -H "Content-Type: application/json" -d '{   "data": {
    "endpoint": "price",
    "base": "XEUX",
    "quote": "EUR", "overrides": { "blocksize-capital": {"XEUX": "ETH"} } 
  }
}'
{"result":1826.8472848996678,"data":{"result":1826.8472848996678},"timestamps":{"providerDataStreamEstablishedUnixMs":1742544459255,"providerDataReceivedUnixMs":1742544459370,"providerIndicatedTimeUnixMs":1742544458872},"statusCode":200,"meta":{"adapterName":"BLOCKSIZE_CAPITAL","metrics":{"feedId":"{\"base\":\"eth\",\"quote\":\"eur\"}"}}}
``` 

[DF-21153]: https://smartcontract-it.atlassian.net/browse/DF-21153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -176,3 +176,34 @@ export const groupArrayByKey = <T extends Record<string, string>, K extends keyo
     {} as Record<T[K], T[]>,
   )
 }
+
+/**
+ * Returns adapter name in lower case and with hyphens replaced by underscores.
+ * Types are declared such that the output type is defined iff the input type
+ * is defined.
+ * @param  name - The name to canonicalize
+ * @returns The canonical name
+ */
+export function getCanonicalAdapterName(name: string): string
+export function getCanonicalAdapterName(name: undefined): undefined
+export function getCanonicalAdapterName(name: string | undefined): string | undefined {
+  return name?.toLowerCase().replace(/-/g, '_')
+}
+
+type StringKeys<T> = { [key: string]: T }
+
+/**
+ * Returns the record with adapter names as keys after canonicalizing the keys.
+ * @param  obj - The record with adapter names as keys
+ * @returns The record with canonicalized adapter names as keys
+ */
+export const canonicalizeAdapterNameKeys = <T>(
+  obj: StringKeys<T> | undefined,
+): StringKeys<T> | undefined => {
+  if (obj === undefined) {
+    return undefined
+  }
+  return Object.fromEntries(
+    Object.entries(obj).map(([key, value]) => [getCanonicalAdapterName(key), value]),
+  )
+}
--- a/test/overrides.test.ts
+++ b/test/overrides.test.ts
@@ -248,7 +248,7 @@ test('adapter name in overrides can have hyphen', async (t) => {
     base: 'OVER2',
     quote: 'USD',
     overrides: {
-      "test-adapter": {
+      'test-adapter': {
         OVER2: 'qweqwe',
       },
     },
--- a/test/overrides.test.ts
+++ b/test/overrides.test.ts
@@ -259,3 +259,20 @@ test('adapter name in overrides can have hyphen', async (t) => {
     quote: 'USD',
   })
 })
+
+test('adapter name in overrides can use upper case', async (t) => {
+  const response = await t.context.testAdapter.request({
+    base: 'OVER2',
+    quote: 'USD',
+    overrides: {
+      TEST_ADAPTER: {
+        OVER2: 'qweqwe',
+      },
+    },
+  })
+
+  t.deepEqual(response.json().data, {
+    base: 'qweqwe',
+    quote: 'USD',
+  })
+})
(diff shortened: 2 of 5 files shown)


src/util/index.ts
```diff
@@ -176,3 +176,34 @@ export const groupArrayByKey = <T extends Record<string, string>, K extends keyo
...
+ * Types are declared such that the output type is defined iff the input type
+ * is defined.
+ * @param  name - The name to canonicalize
+ * @returns The canonical name
+ */
+export function getCanonicalAdapterName(name: string): string
+export function getCanonicalAdapterName(name: undefined): undefined
+export function getCanonicalAdapterName(name: string | undefined): string | undefined {
+  return name?.toLowerCase().replace(/-/g, '_')
+}
```
alecgard, line 191: Could we just do:

```suggestion
export function getCanonicalAdapterName(name?: string): string | undefined {
  return name?.toLowerCase().replace(/-/g, '_')
}
```

test/overrides.test.ts
```diff
@@ -137,7 +137,7 @@ test('request overrides take precedence over adapter hardcoded ones', async (t)
     base: 'OVER1',
     quote: 'USD',
     overrides: {
-      test: {
+      test_adapter: {
```
alecgard, line 140: Can we also test uppercase override names (exact match the adapter name)?
```suggestion
      TEST_ADAPTER: {
```

alecgard (changes requested):

src/util/index.ts
dskloetc: No, I tried to explain this with the comment above:
```
 * Types are declared such that the output type is defined iff the input type
 * is defined.
 ```
but evidently I failed.

If I have `name` of type `string`, then with your version `getCanonicalAdapterName(name)` is of type `string | undefined` which can cause problems if it's passed to something that requires `string`.
With my version `getCanonicalAdapterName(name)` is of type `string` if `name` is of type `string`.

Any suggestion for how I can make this clearer while staying concise?

test/overrides.test.ts
dskloetc: Added an extra test at the bottom.

src/util/index.ts
alecgard: Ahh I missed this comment - that does concisely explain why! One potential alterntive is to use generics (e.g. `T extends string | undefined`), but I wouldn't recommend that as I think it adds more confusion that it solves when defining the return type.

What you've currently got is good!

alecgard (approved):

PR #448, opened 2025-03-24
Fix flaky test with fake clock

dskloetc:
# Description

I noticed a test which flaked because it assumes that `new Date().getTime()` returns the same value throughout the test.

# Changes

Use a fake clock in the test so it doesn't change during the test.


alecgard (approved):
Nice find!

PR #481, opened 2025-05-08
Merc 6849 llo benchmarking on staging mainnet don

namikmesic:


src/util/fast-serialize.ts
```diff
@@ -0,0 +1,204 @@
...
+ *
+ * This implementation uses specialized object detection and hand-optimized
+ * JSON string construction for common response patterns, avoiding the overhead
+ * of the full JSON.stringify process. For complex or unusual responses, it
+ * falls back to standard JSON.stringify.
+ *
+ * @param response - The response object to serialize
+ * @returns JSON string representation of the response
+ */
+export function serializeResponse<T extends ResponseGenerics>(
```
alejoberardino, line 24: I'd be really curious to A/B test this feature, as the JSON.stringify call even tho slow it's using a c lib iirc, but probably for small payloads the string concatenation is faster? It seems somewhat brittle to me though, it's good to try but the approach should be stricter if it turns out it gives significant benefits

src/util/requester.ts
```diff
@@ -81,16 +78,28 @@ interface RequesterResult<T> {
...
 
 /**
- * Centralized management of outbound http requests.
- * Enforces rate limiting on a single instance (complying with the N Readers - 1 Writer arch for EA scaling)
- * by adding requests into a queue, processing them sequentially and sleeping when it reaches its limit.
- * The queue will throw an error if the Requester attempts to add more items than the max configured.
- * It additionally serves to coalesce requests by utilizing a more complex queue structure:
- *   - ignores duplicate items via a provided key
- *   - doesn't use the request itself because it's common for those to have things like timestamps/nonces
+ * Manages outbound HTTP requests with queueing, rate limiting, and connection reuse.
```
alejoberardino, line 81: This one I would honestly remove entirely. It adds complexity but it's only useful for outbound http requests, and we're not really doing that many. Most LL adapters use a streaming protocol anyways, so this should not affect performance at all.

src/validation/input-params.ts
```diff
@@ -1,32 +1,20 @@
...
 import { AdapterError } from './error'
 
-/**
- * Type for the overrides object that is either hardcoded in the adapter,
- * or present in the incoming request body
- */
+/* -------------------------------------------------------------------------- */
+/*                               TYPE HELPERS                                 */
+/* -------------------------------------------------------------------------- */
+
```
alejoberardino, line 10: I would DEFINITELY not remove all these comments haha, I added them with a lot of love and I'm sure our AI overlords appreciate them even if they hallucinate to remove them

src/validation/input-params.ts
```diff
@@ -330,85 +234,73 @@ class ProcessedParam<const T extends InputParameter = InputParameter> {
...
       ([name, param]) => new ProcessedParam(name, param),
     )
-
-    // Check that all options match param type
-    // Check that defaults matches param validation
     this.validateDefinition()
+
+    /* New: compile schema once */
+    this.schema = definitionToJsonSchema(this.definition)
+    this.validateFn = getValidator(this.schema)
```
alejoberardino, line 291: We tried ajv fast too, but at the time the parameters were very complex and proved impossible to do the migrations properly so it was quickly rejected. Still, it's a big time sink on the read request execution, so any speedup helps a lot and worth a try. I would first consider the approach where the validation is skipped entirely for mercury jobs though, as we avoid this altogether. Paired with the suggestion to precalculate the cacheKeys, requests could be resolved in 2 lines only and be blazing fast

alejoberardino (commented):
Potential avenues to take for improvements:

- The redis cache implementation that EAs use have a local in-memory cache, as a sort of layered cache implementation to not have to read from redis all the time. This has evolved since the first implementations, but you can see that it [returns directly](https://github.com/smartcontractkit/ea-framework-js/blob/main/src/cache/redis.ts#L61) if the values are found in memory. Some things here: 
  - Looking at the metrics however I see that the adapter makes a lot of get requests to redis anyways, and I would expect much fewer if the in-memory cache was working as intended. Fixing this would be a huge win for timeliness of requests.
  - Redis is also used as a "subscription set", which serves as the layer in between read and write to let the adapter know what values to fetch. This should also have an in-memory layer, because the zadd ops to add to this set are also hundreds per second and they are not necessary: the entries will fall from the set after a while, probably minutes. This should be another huge win.

alejoberardino:
Another thing to consider is parallelizing multiple read instances and a write instance in a single dockerfile, should be doable with some weird constructs and would not require any changes to the interfaces

PR #492, opened 2025-05-19
Update dependency eventsource from 2.0.2 to 4.0.0 and remove @types/eventsource

dskloetc:
# Description

The [renovate PR](https://github.com/smartcontractkit/ea-framework-js/pull/347) to update `@types/eventsource` to v3 had failures.
While investigating I found that @types/eventsource was [removed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/71870) because the types are included in `eventsource` starting from version 3.0.0.

# Changes

1. Ran `yarn remove @types/eventsource`.
2. Ran `yarn add eventsource`.
3. Fixed import statements.
4. Changed `EventSource.EventSourceInitDict` to `EventSourceInit`
5. Removed `^` from version after `pinned-dependencies.yaml` failed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "ajv": "8.17.1",
     "axios": "1.9.0",
-    "eventsource": "^4.0.0",
+    "eventsource": "4.0.0",
     "fastify": "5.3.3",
     "ioredis": "5.6.1",
     "mock-socket": "9.3.1",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "ajv": "8.17.1",
     "axios": "1.9.0",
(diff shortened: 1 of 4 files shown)


dskloetc:
I looked but I couldn't find any EA using this. I searched for `SseTransport`. The only thing I found was `oanda` but it creates `ModifiedSseTransport` which is not related to `SseTransport`.

Which EA is using this and how did it break for you?

mxiao-cll:
Yes that `oanda` EA, I don't quite remember the details but I think it was some types breaking. As long as we can confirm `oanda` is still good we should be good.

mxiao-cll (approved):

dskloetc:
I did test Oanda with the framework changes and it appeared unaffected.

PR #515, opened 2025-07-08
Add status endpoint

Fletch153:
- Added new endpoint to return system status as described here https://smartcontract-it.atlassian.net/browse/DF-21286
- Status includes redacted configurations, transports, system info etc..
- This information can be used by core node for additional debugging 
- Endpoint results look as such:

```{
  "runtime": "t",
  "nodeVersion": "v20.11.0",
  "platform": "darwin",
  "architecture": "arm64",
  "hostname": "MB-XD96T7W1WN",
  "metrics": {
    "enabled": true,
    "port": 9080,
    "endpoint": "/metrics"
  },
  "adapter": {
    "name": "CFBENCHMARKS",
    "version": "2.4.12",
    "uptimeSeconds": 2.602806208,
    "endpoints": [
      {
        "name": "crypto",
        "aliases": [
          "values",
          "price"
        ],
        "transports": [
          "rest",
          "restsecondary",
          "ws",
          "wssecondary"
        ]
      },
      {
        "name": "birc",
        "aliases": [],
        "transports": [
          "default_single_transport"
        ]
      },
      {
        "name": "crypto-lwba",
        "aliases": [
          "crypto_lwba",
          "crypto lwba"
        ],
        "transports": [
          "default_single_transport"
        ]
      }
    ],
    "defaultEndpoint": "crypto",
    "configuration": [
      {
        "name": "API_ENDPOINT",
        "type": "string",
        "description": "The default REST API base url",
        "required": false,
        "default": "https://www.cfbenchmarks.com/api",
        "customSetting": true,
        "value": "https://www.cfbenchmarks.com/api"
      }
    ]
  }
}```

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,142 +1,167 @@
-import typescriptEslint from "@typescript-eslint/eslint-plugin";
-import tsdoc from "eslint-plugin-tsdoc";
-import tsParser from "@typescript-eslint/parser";
-import path from "node:path";
-import { fileURLToPath } from "node:url";
-import js from "@eslint/js";
-import { FlatCompat } from "@eslint/eslintrc";
+import typescriptEslint from '@typescript-eslint/eslint-plugin'
+import tsdoc from 'eslint-plugin-tsdoc'
+import tsParser from '@typescript-eslint/parser'
+import path from 'node:path'
+import { fileURLToPath } from 'node:url'
+import js from '@eslint/js'
+import { FlatCompat } from '@eslint/eslintrc'
 
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
 const compat = new FlatCompat({
-    baseDirectory: __dirname,
-    recommendedConfig: js.configs.recommended,
-    allConfig: js.configs.all
-});
+  baseDirectory: __dirname,
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
+})
 
-export default [{
+export default [
+  {
     ignores: [
-        "**/node_modules",
-        "**/dist",
-        "**/coverage",
-        "scripts/generator-adapter",
-        "**/.yarn",
-        "**/.vscode",
+      '**/node_modules',
+      '**/dist',
+      '**/coverage',
+      'scripts/generator-adapter',
+      '**/.yarn',
+      '**/.vscode',
     ],
-}, ...compat.extends("eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"), {
+  },
+  ...compat.extends('eslint:recommended', 'plugin:@typescript-eslint/recommended', 'prettier'),
+  {
     plugins: {
-        "@typescript-eslint": typescriptEslint,
-        tsdoc,
+      '@typescript-eslint': typescriptEslint,
+      tsdoc,
     },
 
     languageOptions: {
-        parser: tsParser,
+      parser: tsParser,
     },
 
     rules: {
-        "tsdoc/syntax": "warn",
+      'tsdoc/syntax': 'warn',
 
-        "array-callback-return": ["error", {
-            allowImplicit: false,
-            checkForEach: true,
-        }],
+      'array-callback-return': [
+        'error',
+        {
+          allowImplicit: false,
+          checkForEach: true,
+        },
+      ],
 
-        "no-constant-binary-expression": "error",
-        "no-constructor-return": "error",
-        "no-duplicate-imports": "error",
-        "no-promise-executor-return": "error",
-        "no-self-compare": "error",
-        "no-template-curly-in-string": "error",
-        "no-unmodified-loop-condition": "error",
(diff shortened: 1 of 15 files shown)

### CODE REVIEW

File: src/status/router.ts [hunk 1]
```diff
@@ -0,0 +1,89 @@
...
+      })),
+      defaultEndpoint: adapter.defaultEndpoint,
+      configuration: buildSettingsList(adapter),
+      runtime: {
+        nodeVersion: process.version,
+        platform: process.platform,
+        architecture: process.arch,
+        hostname: hostname()
+      },
+      metrics: {
```
mxiao-cll, line 80: Why not dump the entire adapter.config.settings

Fletch153: We do a few lines up on 73

mxiao-cll (dismissed):

akuzni2 (approved):

alejoberardino (approved):

eslint.config.mjs
dskloetc, line 1: Does the reformatting need to be in this PR?
It makes it a lot harder to see what has changed.
Would it be possible to reformat without changes in a separate PR so that this PR doesn't have unrelated changes?

test/_force-exit.mjs
dskloetc, line 1: I don't see any functional changes in this file? Does it need to be in this PR?

test/status-endpoint.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  if (!statusResponse?.body) {
+    t.fail()
+    return
+  }
+
+  const parsedResponse = statusResponse?.json() as StatusResponse
+
+  // Debug: Print response except for configs
+  const debugResponse = { ...parsedResponse }
+  delete (debugResponse as Partial<StatusResponse>).configuration
```
dskloetc, line 56: This looks unused?

test/status-endpoint.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  // Test runtime information
+  t.is(typeof parsedResponse.runtime.nodeVersion, 'string')
+  t.is(typeof parsedResponse.runtime.platform, 'string')
+  t.is(typeof parsedResponse.runtime.architecture, 'string')
+  t.is(typeof parsedResponse.runtime.hostname, 'string')
+
+  // Test metrics configuration
+  t.is(typeof parsedResponse.metrics.enabled, 'boolean')
+  t.is(parsedResponse.metrics.port, undefined)
+  t.is(parsedResponse.metrics.endpoint, undefined)
```
dskloetc, line 86: Would it be possible to do a deepEquals expectation on the entire `parsedResponse`? That way you don't miss it if additional fields are added.

test/status-endpoint.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  const statusResponse = await api?.inject({ path: '/status' })
+
+  if (!statusResponse?.body) {
+    t.fail()
+    return
+  }
+
+  const parsedResponse = statusResponse?.json() as StatusResponse
+
+  // Test that sensitive API_KEY setting is redacted
```
dskloetc, line 125: Can we test that all the sensitive settings are redacted? Not just this particular one.

test/status-endpoint.test.ts
Fletch153: Some values are environment dependent - uptime, nodeVersion, host name etc.. Also, you could delete the entire response object and the test would still pass. We need to make sure these fields are there

test/status-endpoint.test.ts
```diff
@@ -0,0 +1,139 @@
...
+  t.is(parsedResponse.metrics.port, undefined)
+  t.is(parsedResponse.metrics.endpoint, undefined)
+
+  await api?.close()
+})
+
+test.serial('status endpoint redacts sensitive configuration values', async (t) => {
+  // Set up sensitive environment variables that the framework uses
+  process.env['API_KEY'] = '12312341234'
+  process.env['CACHE_REDIS_PASSWORD'] = 'super-secret-redis-password'
```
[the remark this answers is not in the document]
dskloetc: If somebody accidentally removes `sensitive: true` from `TLS_PRIVATE_KEY`, is there a test that fails?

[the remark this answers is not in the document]
Fletch153: I don't think so as the env variables can be overridden or extended at run time. We can provide sound defaults and test for those but i'm not sure if hardcoding a test to ensure a hardcoded default value is set makes sense. Either way, i would say that test would be beyond the scope of this PR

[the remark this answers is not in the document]
Fletch153: My point is, what should the test be catching?

- Adding a new variable incorrectly as sensitive will be missed
- Adding a new sensitive variable will cause tests to break, as they need to add it again to the expected sensitive list within the test
- If someone is intentionally setting sensitive to false, then why wouldn't they also set the test to false?

The only thing which it might be useful for is someone accidentally setting sensitive to false, then it acts as a double check by forcing the user to also update the test. Which seems like a very unlikely scenario at the expense of needing to maintain 2 sensitive lists.

Anyway, i have added the test to unblock, i don't think it's going to cause much harm

PR #547, opened 2025-08-25
[DF-21594]update the ws reconnect logic and adding explicit logging

guandali:
related to reported bug here: [DF-21594 icap EAs appears hung over the weekend](https://smartcontract-it.atlassian.net/browse/DF-21594?atlOrigin=eyJpIjoiODZiZThjMzI0N2JlNDM1MTlkNTg0NDEyNGRiZWFkMjciLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

[Slack discussion](https://chainlink-core.slack.com/archives/C06H4Q3BNTC/p1753974415484149)

Some identified issue: 
- `lastMessageReceivedAt` is only updated when there a write to cache (if the source sends in heartbeat,  the `lastMessageReceivedAt` ) and that would lead to periodic closure of web socket over the non-market hour (every 120 seconds). 
-  immediate close of ws and reconnect: if we rendered the websocket is stale -> we would call `close` on ws right away and then start a new ws; if the previous instance is not properly closed, we could end up with two instances of ws for same data source and it is known issue for ICAP/TP has two concurrent connections [Context: TP and ICAP EAs use the same credentials, and often there are issues with the set of credentials not having concurrent (ie: 2+) connections enabled.](https://github.com/smartcontractkit/external-adapters-js/tree/main/packages/sources/icap)

What changed in this PR 
- update `lastMessageReceivedAt` whenever receives an update 
- ensure proper close of active ws before proceed to open a new one.


src/transports/websocket.ts
```diff
@@ -336,22 +347,18 @@ export class WebSocketTransport<
...
-    const timeSinceLastMessage = Math.max(0, now - this.lastMessageReceivedAt)
-    const timeSinceConnectionOpened = Math.max(0, now - this.connectionOpenedAt)
-    const timeSinceLastActivity = Math.min(timeSinceLastMessage, timeSinceConnectionOpened)
-    const connectionUnresponsive =
-      timeSinceLastActivity > 0 &&
-      timeSinceLastActivity > context.adapterSettings.WS_SUBSCRIPTION_UNRESPONSIVE_TTL
+   const timeSinceLastMessage = Math.max(0, now - this.lastMessageReceivedAt)
+   const timeSinceConnectionOpened = Math.max(0, now - this.connectionOpenedAt)
+  // Explicit initial grace (1s) after open; after that, rely on "time since last message"
+  const INITIAL_GRACE_MS = 1000
```
mxiao-cll, line 353: Put this somewhere else

src/transports/websocket.ts
```diff
@@ -372,13 +379,23 @@ export class WebSocketTransport<
...
           `Connection was opened only ${timeSinceConnectionOpened}ms ago, waiting for that to get to 1s before continuing...`,
         )
-        await sleep(1000 - timeSinceConnectionOpened)
+        await sleep(INITIAL_GRACE_MS - timeSinceConnectionOpened)
+      }
+      if (this.wsConnection) {
+        await new Promise<void>((resolve) => {
+          this.wsConnection!.once('close', () => resolve())
+          try {
+            this.wsConnection!.close(1000)
```
mxiao-cll, line 392: Do you want to use INITIAL_GRACE_MS here as well?

guandali: its a standard code for normal closing of ws

src/transports/websocket.ts
```diff
@@ -504,3 +507,40 @@ export class WebsocketReverseMappingTransport<
...
 }
+
+/**
+ * Waits for a WebSocket connection to close gracefully.
+ *
+ * @param ws - The WebSocket instance (from the `ws` package).
+ * @param gracefulMs - Max time in milliseconds to wait for a graceful close before terminating. Defaults to 1500 ms.
+ * @returns A promise that resolves when the socket has closed or been terminated.
+ */
+async function waitForGracefulClose(ws: WebSocket, gracefulMs = 1500): Promise<void> {
```
mxiao-cll, line 518: Would be good to have some retry here?

cawthorne:
It seems reasonable to me. We should make sure this is tested on Staging before it goes to Production.

PR #661, opened 2026-01-20
Add WebSocket failover counter metric and URL change logging

cawthorne:
## Summary
Adds observability for WebSocket failover mechanism to help diagnose connection issues.

## Problem
During a Tiingo incident (2026-01-13 03:19-03:32 UTC), we could not determine if failover triggered:
- `streamHandlerInvocationsWithNoConnection` counter not exposed as metric
- Counter increments logged at TRACE level (not visible with LOG_LEVEL=info)
- URL changes logged at DEBUG level and suppressed with `CENSOR_SENSITIVE_LOGS=true`

This made it impossible to answer:
- Did failover trigger during the incident?
- What was the counter value at any given time?
- When did URL switches occur?

### 1. New Prometheus Metric
- Added `ws_connection_failover_count` gauge metric
- Exposes `streamHandlerInvocationsWithNoConnection` value in real-time
- Labeled by `transport_name` for per-transport tracking
- Updated when unresponsive connections are detected


src/transports/websocket.ts
```diff
@@ -431,9 +432,10 @@ export class WebSocketTransport<
     // Check if we should close the current connection
     if (!connectionClosed && (urlChanged || connectionUnresponsive)) {
       if (urlChanged) {
+        logger.info('Websocket URL has changed, closing connection to reconnect...')
         censorLogs(() =>
           logger.debug(
-            `Websocket url has changed from ${this.currentUrl} to ${urlFromConfig}, closing connection...`,
```
dskloetc: Does this no longer close the connection?

cawthorne: It does, but that string got moved to the info level log, so it is always visible with info level logs.

src/metrics/index.ts
```diff
@@ -357,4 +357,9 @@ export const metrics = new Metrics(() => ({
     help: 'The number of addresses in PoR request input parameters',
     labelNames: ['feed_id'] as const,
   }),
+  wsConnectionFailoverCount: new client.Gauge({
+    name: 'ws_connection_failover_count',
+    help: 'The number of consecutive connection issues (unresponsive/no data, abnormal closures), used to trigger URL failover. Resets to 0 when data flows successfully.',
```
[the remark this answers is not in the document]
cawthorne: I actually already had that removed on the description on my local,  just forgot to push.

cl-efornaciari (approved):

PR #720, opened 2026-03-19
Fix Streaming EA outages when receiving different casing for same asset

alecgard:
Two commits
- 51d2e3f adds tests to reproduce the [bug](https://smartcontract-it.atlassian.net/browse/DF-22956) where requests with different casing for the same asset can cause a Streaming EA to return no data
- 1916a02 implements the fix - not overwriting the subscription if we already have it stored with a different casing. Also added a warning log, as this is a sign that different jobs are using different casing, which is unexpected.

Can see the repro tests failed before the fix:
image

And passed after: 
image

### DIFF
--- a/src/util/subscription-set/expiring-sorted-set.ts
+++ b/src/util/subscription-set/expiring-sorted-set.ts
@@ -32,8 +32,23 @@ export class ExpiringSortedSet<T> implements SubscriptionSet<T> {
   add(value: T, ttl: number, key: string) {
     let node = this.map.get(key)
     if (node) {
+      if (JSON.stringify(value) !== JSON.stringify(node.data.value)) {
+        logger.warn(
+          `Subscription set received a value for key "${key}" that differs from the stored value. ` +
+            `Keeping the original value to avoid unnecessary subscription churn. ` +
+            `This indicates requests are using inconsistent parameter casing - ` +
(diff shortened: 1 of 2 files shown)


cll-gg (approved):

cawthorne (approved):

alejoberardino (changes requested):
@alecgard @cl-efornaciari @cawthorne 

I am against doing solving our current problem (some feeds have conflicts) with a framework change right now in general, as I believe there's better alternatives:
- Changing the jobs is slow, but it's the most performant and also keeps our source of truth the cleanest. This should be done regardless, but does fix the issues we have.
- I believe the implemented EAs themselves cause this issue: many of them erase casing at the wrong stage (e.g. before subscribing) instead of making this apply before keys are resolved (i.e.  as request transforms). That erasure is not performant, and when trying to squeeze out performance for high throughput  like we do for these EAs it becomes very relevant; however, if necessary, we should at least fix the EAs from doing it in the wrong place.
- A framework change increases the surface for risk much more than individual EA changes or job fixes.

That said a framework should serve its users, so if the desire is for case insensitivity I can only urge caution as I do see some currently working cases breaking from the change in expectations. Happy to chat more about this too, I'm not currently maintaining this framework or the individual EAs so if you don't agree let me know or dismiss explicitly

PR #753, opened 2026-04-27
Add fallback request

mxiao-cll:
# OPDTA-6402

We want a single endpoint to have the ability to invoke two transports and return value from 2nd transport if there is no data in the 1st transport.

- Added `fallbackTransport` on the endpoint to specify fallback for a given primary
- Call both transport and prioritize primary
- Include the transport name in the output


src/validation/index.ts
```diff
@@ -77,6 +77,10 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
       req,
       adapter.config.settings,
     )
+    const fallbackTransportName = endpoint.getFallbackTransportNameForRequest(
```
dskloetc: Why call this here already while it's only used 35 lines later?

src/validation/index.ts
```diff
@@ -108,6 +112,12 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
 
     // Now that all the transformations have been applied, all that's left is calculating the cache key
     if (endpoint.cacheKeyGenerator) {
+      if (fallbackTransportName) {
+        errorCatcherLogger.warn(
+          'Fallback transports ignored for endpoints with custom cache key generators',
```
dskloetc: Can we just throw an error right away if both cacheKeyGenerator and `fallbackTransport` are specified to the endpoint?

src/adapter/basic.ts
```diff
@@ -449,6 +450,14 @@ export class Adapter<
     const endpoint = this.endpointsMap[req.requestContext.endpointName]
     const transport = endpoint.transportRoutes.get(req.requestContext.transportName)
 
+    return this.handleSingleTransportRequest(req, replySent, transport)
+  }
+
+  private async handleSingleTransportRequest(
```
dskloetc, line 486: What's the purpose of this split?

src/adapter/endpoint.ts
```diff
@@ -186,6 +188,34 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
     return transportName
   }
 
+  getFallbackTransportNameForRequest(primaryTransportName: string, settings: T['Settings']) {
+    if (!settings.TRANSPORT_FALLBACK_ENABLED || !this.fallbackTransport) {
```
dskloetc, line 201: Why do we need `TRANSPORT_FALLBACK_ENABLED`? Isn't the presence of `fallbackTrasport` in the endpoint enough to indicate that we really want to use the fallback transport?

src/adapter/endpoint.ts
```diff
@@ -186,6 +188,34 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
...
+      logger.trace('TRANSPORT_FALLBACK_ENABLED is false or fallbackTransport is not set')
+      return
+    }
+
+    const fallbackTransportName = this.fallbackTransport.toLowerCase()
+
+    if (!this.transportRoutes.get(fallbackTransportName)) {
```
dskloetc: Can we do this check when the endpoint is constructed rather than for every request?

src/adapter/endpoint.ts
```diff
@@ -186,6 +188,34 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
...
+      throw new AdapterError({
+        statusCode: 400,
+        message: `No fallback transport found for key "${fallbackTransportName}", must be one of ${JSON.stringify(
+          this.transportRoutes.routeNames(),
+        )}`,
+      })
+    }
+
+    if (fallbackTransportName === primaryTransportName.toLowerCase()) {
```
dskloetc: This indicates a problem with the design.
`primaryTransportName` is determined by the request.
But `fallbackTransportName` is configured in the adapter.
Usually transport is either `rest` or `ws`.
Perhaps, `ws` will be the default and `rest` will be the fallback.
But I might still specify `rest` in the request. But just because it's now a fallback, I can't specify it anymore?

I feel like either the fallback transport has to be specified in the request, just like the primary transport, or a fallback has to be a fallback for a specific primary transport. But a global fallback regardless of what the primary transport was seems wrong.

src/validation/index.ts
```diff
@@ -125,14 +135,26 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
...
+
+      if (fallbackTransportName) {
+        req.requestContext.fallback = {
+          transportName: fallbackTransportName,
+          cacheKey: calculateCacheKey({
+            data: req.requestContext.data,
+            adapterName: adapter.name,
+            endpointName: endpoint.name,
+            transportName: fallbackTransportName,
+            adapterSettings: adapter.config.settings,
```
dskloetc: What do you think about declaring `commonCacheKeyParams` and using it twice with only different `transportName`?

src/util/types.ts
```diff
@@ -38,6 +38,12 @@ export type AdapterRequestContext<T> = {
   /** Precalculated cache key used to get and set corresponding values from the cache and subscription sets */
   cacheKey: string
 
+  /** Fallback transport context to use if the primary transport is unable to return data */
+  fallback?: {
```
dskloetc, line 42: I don't see where this is actually used. I see it's being set, but where is it read?

src/adapter/basic.ts
mxiao-cll: Later on we will call this thing twice, once for primary once for fallback

src/adapter/endpoint.ts
mxiao-cll: This is mostly for rolling this feature out, I want to be able to control it other than code

src/adapter/endpoint.ts
mxiao-cll: Made it a mapping of primary to fallback

src/util/types.ts
mxiao-cll: Will be used by next PR

src/adapter/endpoint.ts
```diff
@@ -56,7 +56,14 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
       this.transportRoutes = params.transportRoutes
       this.customRouter = params.customRouter
       this.defaultTransport = params.defaultTransport
-      this.fallbackTransport = params.fallbackTransport?.toLowerCase()
+      this.fallbackTransport = params.fallbackTransport
+        ? Object.fromEntries(
+            Object.entries(params.fallbackTransport).map(([k, v]) => [
+              k.toLowerCase(),
```
dskloetc, line 62: Would it make sense to also canonicalize hyphen and underscore?

src/adapter/endpoint.ts
```diff
@@ -71,6 +78,25 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
...
+    // Validate fallback structure if it's defined
+    Object.entries(this.fallbackTransport || {}).forEach(([primary, fallback]) => {
+      if (primary === fallback) {
+        throw new AdapterError({
+          statusCode: 400,
+          message: `Fallback transport "${fallback}" cannot be the same as primary transport.`,
+        })
+      }
+
+      if (!this.transportRoutes.get(fallback)) {
```
dskloetc: The `primary` should also exist, right?

src/validation/index.ts
```diff
@@ -110,12 +106,18 @@ export const validatorMiddleware: AdapterMiddlewareBuilder =
...
+    )
+
     // Now that all the transformations have been applied, all that's left is calculating the cache key
     if (endpoint.cacheKeyGenerator) {
       if (fallbackTransportName) {
-        errorCatcherLogger.warn(
-          'Fallback transports ignored for endpoints with custom cache key generators',
-        )
+        throw new AdapterInputError({
+          message: 'fallbackTransport not supported for endpoints with cacheKeyGenerator',
```
dskloetc: Can we do this in the constructor instead of here?

src/adapter/basic.ts
```diff
@@ -450,7 +450,39 @@ export class Adapter<
...
+    try {
+      return await this.handleSingleTransportRequest(req, replySent, transport)
+    } catch (primaryError) {
+      const fallbackResult = await fallbackResponsePromise
+      if ('response' in fallbackResult) {
+        return fallbackResult.response
+      }
+
+      throw primaryError
+    }
```
dskloetc, line 483: Would this work if you remove the `.then()` call on line 471?
```
} catch (primaryError) {
  try {
    return await fallbackResponsePromise
  } catch (_fallbackError) {
    throw primaryError
  }
}
```
I think that would be easier to follow.

src/adapter/endpoint.ts
mxiao-cll: No, we have a guard against this
```
    if (name !== DEFAULT_TRANSPORT_NAME && !/^[a-z]+$/.test(name)) {
      throw new Error(
        `Transport name "${name}" is invalid. Names in the AdapterEndpoint transports map can only include lowercase letters.`,
      )
    }
```

alejoberardino (changes requested):
I don't think this is the best way to approach this. In this implementation we're making the incoming requests more complex and adding more blocking paths that could inject artificial delays or error out in ways we're not expecting.
I see this much better suited on the transport layer, specifically only the writer side, and the simplest approach is probably to have a transport that composes two inner transports, only uses one subscription set, and at least at first doesn't try anything too smart and keeps both e.g. the ws connection and a polling to the rest backend where both write to the cache.

mxiao-cll:
> I don't think this is the best way to approach this. In this implementation we're making the incoming requests more complex and adding more blocking paths that could inject artificial delays or error out in ways we're not expecting. I see this much better suited on the transport layer, specifically only the writer side, and the simplest approach is probably to have a transport that composes two inner transports, only uses one subscription set, and at least at first doesn't try anything too smart and keeps both e.g. the ws connection and a polling to the rest backend where both write to the cache.

artificial delays
 - Both requests are being fired at the same time, there would only be a slight to none delay when primary fails

error out
 - We try catch everything, there is no unexpected errors

where both write to the cache
 - I think this is even harder to debug having two things over-write each other. The current setup only kicks in when primary fails, and when it kicks in we have deterministic view that only one transport is in-charge


Also according to Nik

```
Does Finage guarantee that their REST endpoint and WS endpoint would return the same price at the same time for the same ticker?
 - not guaranteed, no
 - but should be the case almost all the time
 - It should not be sufficiently different to cause rubberbanding
```

src/cache/response.ts
```diff
@@ -100,6 +100,7 @@ export class ResponseCache<
       ) {
         response.meta = {
           adapterName: calculateAdapterName(this.adapterName, r.params),
+          transportName,
```
dskloetc, line 103: Should we have a test that tests this is the name of the transport that was actually used?

src/adapter/basic.ts
```diff
@@ -449,6 +450,42 @@ export class Adapter<
...
+    const fallbackReq = {
+      ...req,
+      requestContext: {
+        ...req.requestContext,
+        transportName: fallback.transportName,
+        cacheKey: fallback.cacheKey,
+      },
+    } as AdapterRequest<EmptyInputParameters>
+
+    const fallbackResponsePromise = this.handleSingleTransportRequest(
```
dskloetc, line 467: Can you test (or have you tested) what happens when primary succeeds but the fallback fails?
The dangling failing promise might cause some other error handler to trigger.

mxiao-cll: Looks like we do need the .then and .catch because the error can be thrown immediately.

Added more test cases around this

alejoberardino:
> Both requests are being fired at the same time, there would only be a slight to none delay when primary fails
> error out

Still a delay that is introduced that would otherwise not exist

> We try catch everything, there is no unexpected errors

Sorry if it wasn't clear, wasn't talking about exceptions, but any kind of unexpected behavior as a cause of this change, potential race conditions, locks, etc. I usually also like to think that these framework changes may be used in wildly different ways in the actual EA implementations that could break our assumptions

> I think this is even harder to debug having two things over-write each other. The current setup only kicks in when primary fails, and when it kicks in we have deterministic view that only one transport is in-charge

Using the transports would maintain the same pattern debugging-wise that adapters already have, and by manipulating at the reader side you're still hiding the _why_ data was not available on the primary transport (maybe more?).

On a different topic, just thought of the Streams Go images and I think this may break behavior for them due to how they're implemented, but not super sure

dskloetc (approved):
Approving, but please make sure Alejo's concerns are addressed or not a concern.

mxiao-cll:
I still think this is the right solution under the current PMAI analysis.

However, I did another analysis on the incident and in-validated the current PMAI. The conclusion we have is that `We don’t trust either their ws or http server, any of them can go down and we can’t tell when it happens`

Given this new analysis, duel write is prob the way to go

PR #755, opened 2026-04-29
Optional sleep in http transport

mxiao-cll:
Add ability to sleep between http background execute cycles.

One of the use case is fallback transport, we may not want the fallback transport to go full speed on fetching values.

### DIFF
--- a/src/transports/http.ts
+++ b/src/transports/http.ts
@@ -93,6 +93,9 @@ export interface HttpTransportConfig<T extends HttpTransportGenerics> {
     res: AxiosResponse<T['Provider']['ResponseBody']>,
     adapterSettings: T['Settings'],
   ) => ProviderResult<T>[]
+
+  // Optional custom sleep time for the background execute loop
+  backgroundSleepMs?: number
 }
 
 /**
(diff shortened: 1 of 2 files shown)


alejoberardino (changes requested):
This should not be necessary, see https://github.com/smartcontractkit/ea-framework-js/pull/753#pullrequestreview-4193599337

dskloetc:
Can you explain why the fallback transport would go full speed?

mxiao-cll:
> Can you explain why the fallback transport would go full speed?

Fallback transport would go full speed because currently there is no difference between how a fallback and a primary is invoked. Let's say you have a 1 endpoint for rest and 1 endpoint with ws as primary and rest as fallback. The 2 rest endpoint invoked by the 2 endpoints will behave exactly the same.

Given how fallback is intended to be only kick in after primary fail, I think it is ok for it go a little bit slower to save some resources.

dskloetc:
Wouldn't the rate limiter make sure it doesn't go full speed?
The rate limiter affects all transports, not just the fallback. But as far as I can tell the same is true for your background sleep.

mxiao-cll:
> Wouldn't the rate limiter make sure it doesn't go full speed? The rate limiter affects all transports, not just the fallback. But as far as I can tell the same is true for your background sleep.

Rate limiter would effect all transports, if the back-up transport is being used as a primary for a different feed, we don't want that to be effected. So I am looking for a lever that can adjust speed across different endpoints.

For background sleep this is specified on the endpoint level, so I can say for this endpoint sleep and for that don't sleep.

dskloetc (approved):

mxiao-cll:
https://github.com/smartcontractkit/ea-framework-js/pull/753#issuecomment-4391046081

PR #765, opened 2026-05-07
Implement composite cache

mxiao-cll:
Background: https://smartcontract-it.atlassian.net/browse/OPDATA-6402

Commit 1:
- Split `ResponseCache` into `ResponseCache` (base) and `SimpleResponseCache` (implementation)
- Move constructor into base
- Move writeTTL into base
- Move cache entry creation logic into base
- Everything else is in implementation

Commit 2:
 - Re-structure cache so that we can use different transport name for cache key & meta in EA response field
    - If we have a composite transport holding http and ws transport, we still want http and ws transport to write into the same cache key, however the EA response should reflect which transport is used to generate this answer
 - Implement compare cache
   - It holds a responseCache which actually writes into cache
   - It also takes a shouldUpdate function to determine if new entry should be written into cache or not
   - Instead of always loading from cache, it maintains a list of recently written entry so we do a local comparison first
   - It uses responseCache's transport for cache key but it's own transport for EA response meta
 - Implement composite transport
   - Initialize all the child transport with the same compare cache
   - Simply forward each operation to the individual child transport and let cache do the work

Commit 3:
 - Add docs

Commit 4 & 5:
 - Address some comments

Commit 6:
 - Raise an explicit error when sharing transport, currently it just silently fail

Commit 7:
 - Hide composite transport from end user

How to use this: https://github.com/smartcontractkit/external-adapters-js/pull/4939/changes


alejoberardino (commented):
@mxiao-cll from the ticket for this change I agree with your challenge arguments wholeheartedly, would be good to see the intended effect of this change in the concrete impl you have in mind even as a draft

mxiao-cll:
> @mxiao-cll from the ticket for this change I agree with your challenge arguments wholeheartedly, would be good to see the intended effect of this change in the concrete impl you have in mind even as a draft

Big PR here you go and example usage here: https://github.com/smartcontractkit/external-adapters-js/pull/4939/changes

src/cache/response-cache/compare.ts
```diff
@@ -0,0 +1,88 @@
...
+    await this.responseCache.writeEntries(entries)
+
+    entries.forEach(({ key, value }) => {
+      this.localCache.set(key, value)
+    })
+  }
+
+  async writeEntries() {
+    throw new Error('Use write instead for CompareResponseCache')
+  }
```
alejoberardino: this one seems to be internal, can probably just be a method of the simple class and not of this one instead of being part of the interface

src/transports/composite.ts
```diff
@@ -0,0 +1,69 @@
...
+  async registerRequest(
+    req: AdapterRequest<TypeFromDefinition<T['Parameters']>>,
+    adapterSettings: T['Settings'],
+  ): Promise<void> {
+    await Promise.all(this.transports.map((t) => t.registerRequest?.(req, adapterSettings)))
+  }
+
+  async backgroundExecute(context: EndpointContext<T>): Promise<void> {
+    await Promise.all(this.transports.map((t) => t.backgroundExecute?.(context)))
+  }
```
alejoberardino, line 71: promise.all will short circuit failures on any of the transports, for the intended use as a fallback .allSettled seems like a more robust approach, and we log errors for either if they happen while keeping things alive

test/transports/composite.test.ts
```diff
@@ -0,0 +1,121 @@
...
+})
+
+test.afterEach(async (t) => {
+  axiosMock.resetHandlers()
+  t.context.clock.reset()
+  await t.context.testAdapter?.api.close()
+})
+
+test.serial(
+  'composite transport merges child writes using shouldUpdate when run under an adapter',
```
alejoberardino: would be nice to add a couple more test cases for the unhappy paths, e.g. having one transport fail to produce a value, requests coming with a specific transport request, etc

src/cache/response-cache/compare.ts
mxiao-cll: The problem is that the compareCache takes the interface as input instead of the simple class

test/transports/composite.test.ts
mxiao-cll: Added 2 tests, not sure about "requests coming with a specific transport request", that doesn't seem to be related to the new transport added.

src/cache/response-cache/compare.ts
alejoberardino: seems more correct to me to have this expect an extended interface (or even the class) than to have this fail at runtime

src/transports/composite.ts
```diff
@@ -0,0 +1,87 @@
...
+    await Promise.all(
+      Object.entries(this.config.transports).map(([name, transport]) =>
+        transport.initialize(
+          { ...dependencies, responseCache: compareCache },
+          adapterSettings,
+          endpointName,
+          name,
+        ),
+      ),
+    )
```
alejoberardino, line 44: One thing that came to mind when looking at the example usage PR in the ea monorepo (thanks for that btw, made this easier to reason even if very simple) is that we could mistakenly create multiple instances of the same adapter and initialize both, and potentially do unexpected things from the EAs like creating multiple connections to providers when we only intended/support one. 

We need to add something on the framework side to check that if we're using a composite transport for a particular endpoint, either the underlying transports are the same as the composite one, we take the composite one and register the underlying ones automatically, etc.

test/transports/composite.test.ts
alejoberardino: Unless we took it out the framework supports specifying which transport to use when requesting (useful when the EA for example has a crypto endpoint and we know in advance a particular asset pair only exists in the `rest` transport and not the `ws`). I don't think this is a feature we can / should remove, and for safety I'd probably have the framework make sure that cases like these are supported, working, and not allowing for multiple instances of the transports that could cause problems (e.g. multiple ws connections, rate limiting failures to the provider)

alejoberardino (changes requested):

src/cache/response-cache/compare.ts
mxiao-cll: I will implement this instead

src/transports/composite.ts
mxiao-cll: My understanding of the above is that you have concerns on the following two conditions


Case 1
```
  transportRoutes: new TransportRoutes<StockEndpointTypes>()
    .register('ws', wsTransport)
    .register('composite', new CompositeTransport(wsTransport)),
```

Here we are re-using the same transport instance, sharing the same connection, this will fail today before this PR because there will be a cache key collision, one of the routes will not be able to get data.

This is currently allowed in the framework, I will make a change to detect it and throw on EA start-up

Case 2
```
  transportRoutes: new TransportRoutes<StockEndpointTypes>()
    .register('ws', wsTransport())
    .register('composite', new CompositeTransport(wsTransport())),
  customRouter: (_req, adapterConfig) => {
    if (adapterConfig.COMPOSITE_ENABLED) {
      return 'composite'
    }
    return 'ws'
  },
```

This is the example from the EA PR, we have two separate ws transport, each hold a different connection. However `customRouter` here prevents both instance from being activated at the same time. Upon EA start up, you either get the composite transport or the ws transport. Even if you manually pass-in "transport" on the request level. This feature allows this whole thing to work, as we can guarantee there is only 1 connection happening.

I don't think any adjustment is needed here.


Is there any other case I didn't cover that you have concern of?

alejoberardino: Reusing the transport instances is what we should be doing imo, otherwise the risk of users running into errors seems very high (setting up things outside of the normal expectations, but still). For example, the transport itself fetching an auth token and having the two instances replace each other's session.

With regards to this causing a cache key collision, regardless of whether we requested to a composite path that will fetch in two transports or we explicitly asked for only one, the resulting appropriate value should be available in both cases. We can make a choice here:

- We support specifying a transport, composite, of falling back to a default on composite routes. This is the path that breaks less expectations about EAs, but is perhaps more difficult to get right.
- We remove this functionality for composite routes, and allow only composite or separate as an EA config. This will prevent unintended behavior, but is more brittle (configs we don't control directly), breaks existing functionality, and will still need framework changes for the sake of ergonomics to avoid misuse. Relying on the customRouter is not enough, since it could easily be forgotten, unintentionally changed, etc.

I have some ideas as to how to potentially tackle this in both scenarios, but it would be good to decide on the above first to avoid unnecessary back and forth (happy to jump on  a call too!)

mxiao-cll: I am gonna resolve all open items and post a comment about next steps

mxiao-cll:
Action items after discussion

1. Create a backlog item to throw an error when the transport used is different from transport specified
For example if in your request you say `"transport": "ws"` but the request ended up landing on `"http"` due to `customRouter` EA should throw an error
See: https://smartcontract-it.atlassian.net/browse/OPDATA-7374
2. Instead of building the transport on the endpoint, use a simple boolean `enableComposite: true` to indicate the framework to build CompositeTransport and enable it, it have providerIndicatedTimeUnixMs logic by default

src/transports/composite.ts
```diff
@@ -0,0 +1,72 @@
...
+    adapterSettings: T['Settings'],
+  ): Promise<void> {
+    const results = await Promise.allSettled(
+      Object.values(this.transports).map((t) => t.registerRequest?.(req, adapterSettings)),
+    )
+    results
+      .filter((r) => r.status === 'rejected')
+      .forEach((r) => {
+        logger.error(`Transport registerRequest failed: ${r.reason}`)
+      })
```
alejoberardino: this is potentially problematic because it may be the case that one transport fails to register but the other doesn't, but I don't have a much better way to handle it in mind (plus probably not likely to fail registry in the first place)

alejoberardino (dismissed):

src/transports/composite.ts
mxiao-cll: What we do here actually doesn't matter since there is a big try catch outside of this.

When a transport fails to register we either
1. Run the ones that do register successfully or
2. Run nothing

I think 1 makes sense as we are treating the other transport as failure here.

I will roll this out and see if any adjustment needs to be made

src/adapter/endpoint.ts
```diff
@@ -96,6 +104,14 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
       responseCache,
     }
 
+    if (this.enableCompositeTransport && adapterSettings.COMPOSITE_TRANSPORT) {
```
dskloetc, line 114: How about logging a warning if `this.enableCompositeTransport` but not `adapterSettings.COMPOSITE_TRANSPORT`?

src/cache/response-cache/compare.ts
```diff
@@ -0,0 +1,94 @@
...
+        continue
+      }
+      const entryInCache = await this.get(key)
+      if (!this.shouldUpdate(value, entryInCache)) {
+        continue
+      }
+      filteredEntries.push({ key, value })
+    }
+
+    await this.responseCache.writeEntries(filteredEntries)
```
dskloetc: Can we skip this if `filteredEntries` is empty?

src/cache/response-cache/compare.ts
```diff
@@ -0,0 +1,94 @@
...
+    const filteredEntries: {
+      key: string
+      value: AdapterResponse<T['Response']>
+    }[] = []
+
+    for (const { key, value } of entries) {
+      if (!this.shouldUpdate(value, this.localCache.get(key))) {
+        continue
+      }
+      const entryInCache = await this.get(key)
```
dskloetc: Isn't this unnecessarily serial? Can we parallelize it? Something like
```
const filteredEntries = Promise.all(entries.map(async ({ key, value }) => {
  if (!this.shouldUpdate(value, this.localCache.get(key))) {
    return undefined
  }
  const entryInCache = await this.get(key)
  if (!this.shouldUpdate(value, entryInCache)) {
    return undefined
  }
  return value
})
  .filter((entry) => entry !== undefined)
```

src/cache/response-cache/compare.ts
```diff
@@ -0,0 +1,94 @@
...
+    super({
+      inputParameters: responseCache.inputParameters,
+      adapterName: responseCache.adapterName,
+      endpointName: responseCache.endpointName,
+      adapterSettings: responseCache.adapterSettings,
+      dependencies: responseCache.dependencies,
+    })
+    this.transportName = transportName
+    this.responseCache = responseCache
+    this.localCache = new Map()
```
dskloetc, line 44: Do we really need a local cache? It seems to me that if we do, then it should be the responsibility of the simple response cache to have a local cache. Not the compare cache. Is there something specific about the compare cache that makes local caching more important than it already was?

src/cache/response.ts
```diff
@@ -1,179 +1,3 @@
...
-        endpointName: this.endpointName,
-        adapterSettings: this.adapterSettings,
-      })
-
-      this.cache.setTTL(key, ttl)
-    }
-  }
-}
+export { ResponseCache } from './response-cache/base'
+export { SimpleResponseCache } from './response-cache/simple'
```
dskloetc, line 2: Have you checked if the name change will break a lot of EA code?

src/transports/abstract/subscription.ts
```diff
@@ -19,13 +19,18 @@ export abstract class SubscriptionTransport<T extends TransportGenerics> impleme
...
   name!: string
+  initialized = false
 
   async initialize(
     dependencies: TransportDependencies<T>,
     adapterSettings: T['Settings'],
     endpointName: string,
     name: string,
   ): Promise<void> {
+    if (this.initialized) {
```
dskloetc, line 30: Is this related to the composite transport or just a drive-by change?

src/transports/composite.ts
```diff
@@ -0,0 +1,72 @@
...
+    req: AdapterRequest<TypeFromDefinition<T['Parameters']>>,
+    adapterSettings: T['Settings'],
+  ): Promise<void> {
+    const results = await Promise.allSettled(
+      Object.values(this.transports).map((t) => t.registerRequest?.(req, adapterSettings)),
+    )
+    results
+      .filter((r) => r.status === 'rejected')
+      .forEach((r) => {
+        logger.error(`Transport registerRequest failed: ${r.reason}`)
```
dskloetc: Wouldn't we want to know here which of the transports failed?

src/transports/composite.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async backgroundExecute(context: EndpointContext<T>): Promise<void> {
+    const results = await Promise.allSettled(
+      Object.values(this.transports).map((t) => t.backgroundExecute?.(context)),
+    )
+
+    results
+      .filter((r) => r.status === 'rejected')
+      .forEach((r) => {
+        logger.error(`Transport backgroundExecute failed: ${r.reason}`)
```
dskloetc: And here also.

src/transports/index.ts
```diff
@@ -7,6 +7,7 @@ import { InputParametersDefinition, TypeFromDefinition } from '../validation/inp
 export * from './http'
 export * from './sse'
 export * from './websocket'
+export * from './composite'
```
dskloetc: Should this be exported here given that it shouldn't be used directly?

src/cache/response-cache/compare.ts
mxiao-cll: The compare cache would need to read from cache every time before it writes. Simple cache simply write it in blindly.

I am putting a Map here so that we can reduce the number of cache reads

src/cache/response.ts
mxiao-cll: No because the name of ResponseCache did not change

src/transports/abstract/subscription.ts
mxiao-cll: Just a drive-by

dskloetc (approved):

PR #776, opened 2026-05-13
Stop sending undefine ws message + pass through raw endpoint name

mxiao-cll:
## Summary
- **`requestContext.rawEndpointName`** — Stores the normalized endpoint string from the request (the same value used to resolve the endpoint in `endpointsMap`). **`requestContext.endpointName`** remains the canonical endpoint name. When the client uses an **alias**, `endpointName` is the canonical name and `rawEndpointName` is the alias. Adapters can distinguish client-facing endpoint identifiers while keeping a single endpoint definition, transport, and WebSocket connection when multiple names map to the same `AdapterEndpoint`.
- **WebSocket transport** — After `subscribeMessage` / `unsubscribeMessage` run, **`undefined` results are filtered out** before serialize/send, so the provider does not receive spurious payloads (e.g. the string `"undefined"`).
## Motivation
Some adapters expose several request-facing names (aliases) on one endpoint that shares one streaming connection. Previously only the canonical name was available on `requestContext`, so transports could not tell whether the caller used e.g. `stock` vs `stock_quotes` when both resolve to the same endpoint. Exposing the lookup string as `rawEndpointName` fixes that for any code path that has the full **`AdapterRequest`**.

Subscribe/unsubscribe builders may intentionally return **`undefined`** when no wire message is required; those must not be sent to the WebSocket server.

### DIFF
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -33,7 +33,7 @@ export type AdapterRequestContext<T> = {
   endpointName: string
 
   /** Name of the endpoint from input before aliasing */
-  rawEndpointName: string
+  requestEndpointName: string
 
   /** Name of the transport this payload should be directed to */
   transportName: string
--- a/src/util/types.ts
+++ b/src/util/types.ts
@@ -32,6 +32,9 @@ export type AdapterRequestContext<T> = {
   /** Name of the endpoint this payload should be directed to */
   endpointName: string
 
+  /** Name of the endpoint from input before aliasing */
+  rawEndpointName: string
+
   /** Name of the transport this payload should be directed to */
   transportName: string
 
--- a/test/transports/websocket.test.ts
+++ b/test/transports/websocket.test.ts
@@ -10,6 +10,7 @@ import {
   WebSocketClassProvider,
   WebsocketReverseMappingTransport,
   WebSocketTransport,
+  type WebSocketTransportConfig,
 } from '../../src/transports'
 import { SingleNumberResultResponse, sleep } from '../../src/util'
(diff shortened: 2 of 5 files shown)


src/util/types.ts
```diff
@@ -32,6 +32,9 @@ export type AdapterRequestContext<T> = {
   /** Name of the endpoint this payload should be directed to */
   endpointName: string
 
+  /** Name of the endpoint from input before aliasing */
+  rawEndpointName: string
```
dskloetc: How about calling it `requestEndpointName` instead?

test/transports/websocket.test.ts
```diff
@@ -200,6 +202,81 @@ test.serial('connects to websocket, subscribes, gets message, unsubscribes', asy
...
+  // ETH first so the socket gets a subscribe + provider message before SKIP (no subscribe payload).
+  await testAdapter.request({ base: 'ETH', quote: 'DOGE' })
+  await runAllUntilTime(t.context.clock, BACKGROUND_EXECUTE_MS_WS + 200)
+  await testAdapter.request({ base: 'SKIP', quote: 'DOGE' })
+  await runAllUntilTime(t.context.clock, BACKGROUND_EXECUTE_MS_WS + 200)
+  await testAdapter.request({ base: 'ETH', quote: 'SKIPUNSUB' })
+  await runAllUntilTime(t.context.clock, BACKGROUND_EXECUTE_MS_WS + 200)
+
+  t.false(
+    messagesReceived.includes('S:SKIP/DOGE'),
```
dskloetc: Could we just set an expectation on the full value of `messagesReceived`? Or is there something unpredictable about it?

dskloetc (approved):

PR #794, opened 2026-05-25
Add address validator

mxiao-cll:
#OPDATA-7291

Add an address validation utility, some code taken from https://github.com/smartcontractkit/external-adapters-js/blob/main/packages/composites/proof-of-reserves/src/utils/addressValidator.ts

We will be moving address validation from PoR into the protocol adapter


src/validation/address.ts
```diff
@@ -0,0 +1,56 @@
...
+import { address as solanaAddress } from '@solana/kit'
+
+export enum AddressType {
+  ETHEREUM,
+  ETHEREUM_WITHDRAWAL_CREDENTIALS,
+  BITCOIN,
+  SOLANA,
+}
+
+export const validateAddress = (address: string | undefined | null, type: AddressType): boolean => {
```
dskloetc: Why wouldn't the caller just call the corresponding function for the given type directly?
Will this be used in a way that makes having the type a parameter more convenient? I.e., will the type not just always be hard coded at the calling site?

src/validation/address.ts
```diff
@@ -1,42 +1,24 @@
...
-export enum AddressType {
-  ETHEREUM,
-  ETHEREUM_WITHDRAWAL_CREDENTIALS,
-  BITCOIN,
-  SOLANA,
-}
+export const validateEthereumAddress = (address?: string | null) => isAddress(address)
 
-export const validateAddress = (address: string | undefined | null, type: AddressType): boolean => {
+export const validateEthereumWithdrawalCredentials = (address?: string | null) => {
```
dskloetc: Nit: I thought the names starting with "is" were clearer. With "validate" you might think the function will throw, but with "is" it's clear they'll return a `boolean`.

dskloetc (dismissed):

dskloetc (approved):

PR #804, opened 2026-06-03
OPDATA-7692 filter endpoints by env var initial commit

mmcallister-cll:
https://smartcontract-it.atlassian.net/browse/OPDATA-7692


test/adapter/filter-endpoints.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { NopTransport } from '../../src/util/testing-utils'
+
+test.before(() => {
+  LoggerFactoryProvider.set()
+})
+
+const makeEndpoint = (name: string) => new AdapterEndpoint({ name, transport: new NopTransport() })
+
+test.afterEach(() => {
+  delete process.env['ENABLED_ENDPOINTS']
```
dskloetc, line 13: Cleanup should go in `beforeEach`. If a test fails to clean up, you want that to fail, not some random next test.

This is more of an issue with jest where you can have before/afterEach apply to subsets of tests, but still a good rule of thumb to follow.

`afterEach` is only for assertions like "no errors happened during the test", or for cleanups that are not allowed to run initially.

src/adapter/basic.ts
```diff
@@ -184,6 +185,47 @@ export class Adapter<
     }
   }
 
+  /**
+   * Filters the adapter's endpoints based on the ENABLED_ENDPOINTS setting.
+   * If ENABLED_ENDPOINTS is not set, all endpoints are loaded.
+   * If ENABLED_ENDPOINTS is set, only the specified endpoints are loaded.
+   * Throws an error if none of the specified endpoints match.
```
dskloetc, line 192: I think it should throw if any of the listed endpoints don't match. That would be a mistake that you want to find out sooner rather than later.

test/adapter/filter-endpoints.test.ts
```diff
@@ -0,0 +1,68 @@
...
+})
+
+test('filters to only the named endpoints', (t) => {
+  process.env['ENABLED_ENDPOINTS'] = 'price'
+  const adapter = new Adapter({
+    name: 'TEST',
+    endpoints: [makeEndpoint('price'), makeEndpoint('volume')],
+  })
+  t.is(adapter.endpoints.length, 1)
+  t.is(adapter.endpoints[0].name, 'price')
```
dskloetc, line 31: Combine these 2 lines into
```
t.deepEqual(adapter.endpoints.map(e => e.name), ['price'])
```
The error will be more useful than just the length not matching.

test/adapter/filter-endpoints.test.ts
```diff
@@ -0,0 +1,68 @@
...
+})
+
+test('trims whitespace and lowercases names in ENABLED_ENDPOINTS', (t) => {
+  process.env['ENABLED_ENDPOINTS'] = ' Price , VOLUME '
+  const adapter = new Adapter({
+    name: 'TEST',
+    endpoints: [makeEndpoint('price'), makeEndpoint('volume')],
+  })
+  t.is(adapter.endpoints.length, 2)
```
dskloetc, line 40: same

PR #805, opened 2026-06-03
OPDATA-6745: Fix memory leak in Requester

dskloetc:
# [OPDATA-6745](https://smartcontract-it.atlassian.net/browse/OPDATA-6745)

# Description

There is a memory leak in `Requester`.

If requests come in frequently enough that the requester queue is never empty, even if it stays at a small size, the recursive `processNext` function will chain `Promise` objects forever. Only when the request queue is empty, does the promise stack unravel to release the memory.

This happens in `por-address-list`, `the-network-firm`, `tiingo` and to a lesser extent `finnhub-secondary` because they have multiple endpoint making multiple requests. This means that even though one endpoint waits for all its requests to finish before sending a new batch, there is always a batch from another endpoint still in the queue to keep the promise chain from unraveling.

This PR fixes the memory leak by refactoring the `Requester`, in particular the queueing process.

Instead of dropping requests in a queue and separately handling the queue, the code is more procedural with each "thread" taking care of its own request and a queue that only coordinates how requests wait on each other to satisfy the rate limiter.
 
# Changes

1. Add `turn-queue.ts`. This coordinates the different "threads" waiting on the rate limited sequentially.
2. Add tests for `TurnQueue`.
3. Refactor `Requester` using the `TurnQueue`.
4. Update 2 tests that were observing bugs and 1 test that seemed to be observing a bug but has behavior we want to keep.

# Testing

Tested locally with `the-network-firm` with all 24 feeds from RDD.
We'll confirm that it fixes the memory leak when we test in staging after merging.

[OPDATA-6745]: https://smartcontract-it.atlassian.net/browse/OPDATA-6745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/util/turn-queue.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+      first.drop()
+      this.active.next = first.next
+    }
+
+    return turn.starting
+  }
+
+  startNext() {
+    this.length--
```
alejoberardino, line 115: Maybe I'm missing it, but I think this queue could grow unbounded since we're never dereferencing old nodes, simply advancing the currently selected active node

dskloetc: I'm not sure I understand what you're saying but on the next line when we do `this.active = this.active.next`, the node that was previously referenced by `this.active` is no longer referenced and will be garbage collected. Does that address your concern?

alejoberardino: ah saw the active but thought there was still a ref backwards, but yeah it will only be from active forwards, disregard ty!

src/util/requester.ts
```diff
@@ -93,8 +41,7 @@ interface RequesterResult<T> {
  *   - Contemplate architectures with multiple writer EA instances
  */
 export class Requester {
-  private processing = false
-  private queue: UniqueLinkedList<QueuedRequest>
+  private queue: TurnQueue
   private map = {} as Record<string, QueuedRequest>
```
mmcallister-cll: should we take the opportunity to rename `QueuedRequest` type to something like `PendingRequest` and `map` to something more meaningful?

src/util/turn-queue.ts
```diff
@@ -0,0 +1,119 @@
...
+  last: TurnNode
+
+  constructor(private maxLength: number) {
+    const ready = new TurnNode(this)
+    ready.isDone = true
+    this.active = ready
+    this.last = ready
+  }
+
+  async run(task: () => Promise<void>): Promise<void> {
```
mmcallister-cll: I see the minimal and straightforward intent behind calling this `run()` and it's alright, but I have 2 readability suggestions:
- the Queue could be a bit more intuitive for me with an explicit `queue.run()` or `queue.start()` init call outside of the constructor creation that at least sets the sentinel `node.isDone = true` to signal that the queue execution has started and it's accepting new nodes (vs implicit now)
- this current `run` function could have a more descriptive name like `enqueueTask`, or `enqueueTaskAndExecuteOnTurn`, or even something like `enqueueAndRunTask`

Anyway, not married to either suggestion but have a quick think please 🙏

src/util/requester.ts
mmcallister-cll: Also, any reason to not make this a Map<>()?

src/util/requester.ts
```diff
@@ -179,87 +79,111 @@ export class Requester {
       return existingQueuedRequest.promise as Promise<RequesterResult<T>>
     }
 
+    // Set configured timeout for all requests unless manually specified
+    req.timeout = req.timeout || this.timeout
```
mmcallister-cll, line 83: 🤔 could we still cause retained memory (practically _not_ unbounded though) if `API_TIMEOUT = 0`, which axios interprets as no timeout (rather than 0ms timeout), and a hung request/promise?

Asked a different way, to avoid an edge case where some memory on a stuck request never gets cleared cause axios never times out, should we set `API_TIMEOUT` minimum value = 1 rather than 0?

src/util/requester.ts
```diff
@@ -179,87 +79,111 @@ export class Requester {
...
-    return this.processNext()
+      if (e instanceof EvictedError) {
+        const overflowedRequest = req
+        censorLogs(() =>
+          logger.debug(
+            `Request (Key: ${overflowedRequest.key}, Retry #: ${overflowedRequest.retries}) was removed from the queue to make room for a newer one (Size: ${this.queue.length})`,
+          ),
+        )
+        metrics.get('requesterQueueOverflow').inc()
+        throw new AdapterRateLimitError({
```
mmcallister-cll, line 152: silly q: this AdapterRateLimitError would be solved by a higher API tier because it would allow us to get through the queue faster, right?

I was debating asking whether a different error type was appropriate here, but i think this is correct. Is there any other reason we could hit this overflow though or solution to suggest other than increase max queue length towards infinity which doesn't really make sense for our practical uses (being constant traffic and not bursts)?

src/util/requester.ts
dskloetc: Done, done and done.

src/util/turn-queue.ts
dskloetc: I changed it to `runInTurn` to make it a bit more obvious that it doesn't just outright run it.

I think `enqueueTask` hides more than it helps (does it actually run or do you need to pop it from the queue manually?).

And `enqueueAndRunTask` feels too long and hard to parse.

I don't think it's useful to create the queue in a state where it doesn't work so I'm not in favor of a separate `start()` function.

src/util/requester.ts
dskloetc: It's a fair question. But it's orthogonal to the issue at hand so I wouldn't do it in this PR.

And it's a breaking change. If anyone currently has `API_TIMEOUT=0` configured, it would fail to restart on update. It's not currently causing any issues as far as I know.

With `HttpTransport` it doesn't send the next batch of requests until all the previous ones have finished. So it would just stop sending requests. With `SubscriptionTransport`, the background handler can time out so after 3 minutes it does send a new batch of requests. In both cases I think you would notice more problems than just memory running out. So I think I'd just leave it the way it is.

src/util/requester.ts
dskloetc: HttpTransport sends the requests for all subscriptions at the same time. Then it waits for all requests to finish before sending more requests. So increasing the rate limit doesn't help. It's the number of subscriptions (so number of served feeds) that counts. The amount of memory used is already proportional to the number of subscriptions so I don't think there's anything to do other than make sure the queue size is bigger than the number of subscriptions.

With `SubscriptionTransport` we could hit it if we never manage to send all requests without the background handler timeout. But we typically don't set a rate limit when we use `SubscriptionTransport` although we do on some EAs that mix `SubscriptionTransport` and `HttpTransport`.

mmcallister-cll: alright, i think we're good in practice here

mmcallister-cll (approved):

PR #806, opened 2026-06-03
OPDATA-6934 add delta and snapshot modes + related linter refactoring…

mmcallister-cll:
…. Also allow numbers in endpoint names

Adds optional options.batchSubscribeMode to the WebSocketBatchBuilders to configure whether to send subscriptions.desired (full snapshot) or subscriptions.new (delta) upon a new subscription being present.

https://smartcontract-it.atlassian.net/browse/OPDATA-6934
To be used in Nobi EA.

### DIFF
--- a/docs/components/transport-types/websocket-transport.md
+++ b/docs/components/transport-types/websocket-transport.md
@@ -142,7 +142,32 @@ builders: {
 }
 ```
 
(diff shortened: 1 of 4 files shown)


dskloetc:
Can you provide more context in the PR description?

From the title and description it sounds like this PR does 4 different things. Is that necessary?

mmcallister-cll:
Closing as this was addressed in a different way in a separate PR

PR #810, opened 2026-06-04
Add generated unit test for generated customBg transport

dskloetc:
Follow-up to https://github.com/smartcontractkit/ea-framework-js/pull/784 and https://github.com/smartcontractkit/ea-framework-js/pull/788

# Description

When generating a new EA with `yarn new`, it includes an integration test but only unit tests for `http` and `ws` transports.
This PR adds a unit test for `customBg` transports.

# Changes

1. In the generated CustomBg transport file:
    1. Use the name of the endpoint in the transport name, is we usually do manually.
    2. Initialize `config` and `endpointName`.
    3. Add logic to use the requester.
2. Add a customBg unit test template.
3. Use the template to generate the unit test.
4. Update the integration test to match the new logic in the generated transport.

# Testing

1. We have a [workflow](https://github.com/smartcontractkit/ea-framework-js/blob/225d0063ac1979f2f1ba150e1e3b29e30c7139fb/.github/workflows/main.yaml#L77) that generates an adapter that still passes.
3. I manually tested it in `external-adapters-js` to create an adapter with a single endpoint called `abc-def` with a single `customBg` transport. The generated tests passed.


scripts/generator-adapter/generators/app/templates/test/unit/custombg.test.ts.ejs
```diff
@@ -0,0 +1,278 @@
...
+    const requestKey = calculateHttpRequestKey<BaseEndpointTypes>({
+      context: {
+        adapterSettings,
+        inputParameters,
+        endpointName,
+      },
+      data: requestConfig.data,
+      transportName,
+    })
+    /*
```
mmcallister-cll: needed? uncomment?

scripts/generator-adapter/generators/app/templates/test/unit/custombg.test.ts.ejs
```diff
@@ -0,0 +1,278 @@
...
+    requester,
+    responseCache,
+    subscriptionSetFactory: {
+      buildSet: jest.fn(),
+    },
+  } as unknown as TransportDependencies<BaseEndpointTypes>)
+
+  let transport: <%= normalizedEndpointNameCap %>Transport
+
+  type RequestConfig = {
```
mmcallister-cll, line 62: should this match the below object?

scripts/generator-adapter/generators/app/templates/test/integration/adapter.test.ts.ejs
```diff
@@ -41,11 +45,15 @@ describe('execute', () => {
         endpoint: '<%= endpoints[i].inputEndpointName %>',
         transport: '<%= transportName %>'
       }
+<% if (mockPostResponse) { %>
```
mmcallister-cll, line 44: qq: Should this be keyed off of the transport type instead of a new param?
If not, then this is OK.

mmcallister-cll (dismissed):
a couple nits but looks fine

scripts/generator-adapter/generators/app/templates/test/unit/custombg.test.ts.ejs
dskloetc: I had it originally but I thought it's too verbose so I split out the `debugLog` and only assert on the non-debug `log`.

Removed.

scripts/generator-adapter/generators/app/templates/test/integration/adapter.test.ts.ejs
dskloetc: This is actually a pretty fundamental question.

Say you have 2 possible situations: A and B. And in situation A you want to do X and in situation B you want to do Y. You can call the parameter that controls what to do `situationB` or `doY`. I think people often start out with a parameter called `situationB` but then over time `situationB` and `doY` start to diverge and you end up with a parameter called `situationB` which actually means `doY`. So I tend to call these things `doY` from the start.

In this case, since we already have `setBgExecuteMsEnv` and `includeHttpFixtures`, that seemed consistent enough.

scripts/generator-adapter/generators/app/templates/test/unit/custombg.test.ts.ejs
dskloetc: Yes, thanks.

This should have been caught by the compiler. It turns out that in `external-adapters-js` it would have been caught by the compiler, but apparently I only ran the test and didn't try to compile it. And I didn't create a PR there to see it pass CI.

In `ea-framework-js` we are apparently missing a check that the generated code compiles. I tried adding it but there are more issues so I will address them in a separate PR.

I did make some additional changes to make compilation pass in `external-adapters-js` and to bring the generated code more in line with how we tend to make subscription transports.

scripts/generator-adapter/generators/app/templates/test/integration/adapter.test.ts.ejs
mmcallister-cll: mmmm yeah makes sense, i just saw the added param that seemed specific to the new transport tests and figured i'd ask about your decision. Thx!

mmcallister-cll (approved):

PR #808, opened 2026-06-03
[Streams adapter] 504 error on view-function-multi-chain adapter

yaroslav-glukhov-chainlink:
Added FEED_ID_JSON support. 
Added sorting for cache key generation to make key/sha1 hash sematically determentstic.

PR #812, opened 2026-06-05
Allow FEED_ID_JSON in makeStub

dskloetc:
# Description

https://github.com/smartcontractkit/ea-framework-js/pull/808 breaks a lot of tests that use `makeStub`.
`makeStub` is a convenient way in tests to check that every field on a stub object that is being accessed is specified in the test.
The above PR added an extra field to `adapterSettings` which is always being accessed, but not specified in any test.
So all tests that use `makeStub` with `adapterSettings` will now fail.
This includes the generated example adapter in this very repo.
So currently CI is failing on every PR.

This wasn't caught on the breaking PR itself, because the generated example adapter uses the published framework version so it only started failing when a new framework version was published.

# Changes

1. Add `FEED_ID_JSON` to the list of fields that `makeStub` is allowed to access even if it's not specified.
2. When testing the adapter generator in CI, use the local framework version such that similar breakages would be caught on CI in the future.


.github/workflows/main.yaml
```diff
@@ -94,3 +94,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           EXTERNAL_ADAPTER_GENERATOR_NO_INTERACTIVE: 'true'
           EXTERNAL_ADAPTER_GENERATOR_STANDALONE: 'true'
+          EXTERNAL_ADAPTER_GENERATOR_USE_LOCAL_FRAMEWORK_VERSION: 'true'
```
mxiao-cll, line 97: When would this be false?

dskloetc: When you generate a real new adapter in `external-adapters-js`.
Then it uses the published framework version.

mxiao-cll (approved):

PR #823, opened 2026-06-10
OPDATA-6402 update comparison in CompositeTransport CompareResponseCache

mmcallister-cll:
https://smartcontract-it.atlassian.net/browse/OPDATA-6402

Updates comparison in CompositeTransport beyond strict new.timestamp > old.timestamp

New rules address:

1. new.timestamp > old.timestamp
2. refreshing cache TTL
3. same timestamp but different value received after a staleness threshold

Introduces `COMPOSITE_TRANSPORT_STALE_THRESHOLD_MS` setting var defaulted to `CACHE_MAX_AGE / 2` for point 3 above.

### DIFF
--- a/src/cache/response-cache/compare.ts
+++ b/src/cache/response-cache/compare.ts
@@ -14,11 +14,6 @@ export class CompareResponseCache<
   readonly transportName: string
   // The actual cache where responses are written to
   responseCache: ResponseCache<T>
-  // A local map to keep track of the most recent entries written to the responseCache
-  // We compare with this first before comparing with value in cache
-  // so that we can reduce cache reads
-  localCache: Map<string, { value: AdapterResponse<T['Response']>; writtenAt: number }>
-  staleThreshold: number
   // True if next should replace current in cache
   shouldUpdate: (
     next: AdapterResponse<T['Response']>,
@@ -32,7 +27,6 @@ export class CompareResponseCache<
       next: AdapterResponse<T['Response']>,
       current?: AdapterResponse<T['Response']>,
     ) => boolean,
-    staleThreshold = Infinity,
   ) {
     super({
       inputParameters: responseCache.inputParameters,
@@ -43,8 +37,6 @@ export class CompareResponseCache<
     })
     this.transportName = transportName
     this.responseCache = responseCache
-    this.localCache = new Map()
-    this.staleThreshold = staleThreshold
     this.shouldUpdate = shouldUpdate
   }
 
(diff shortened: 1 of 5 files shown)


src/transports/composite.ts
```diff
@@ -24,12 +24,30 @@ export class CompositeTransport<T extends TransportGenerics> implements Transpor
...
 
+    const staleThreshold =
+      adapterSettings.COMPOSITE_TRANSPORT_STALE_THRESHOLD_MS ?? adapterSettings.CACHE_MAX_AGE / 2
+
     const compareCache = new CompareResponseCache(
       transportName,
       this.responseCache,
-      (next, current) =>
-        (next.timestamps?.providerIndicatedTimeUnixMs ?? 0) >
-        (current?.timestamps?.providerIndicatedTimeUnixMs ?? 0),
```
mxiao-cll, line 32: Wouldn't this be as simple as 
```
(next.timestamps?.providerIndicatedTimeUnixMs ?? 0) >=
        (current?.timestamps?.providerIndicatedTimeUnixMs ?? 0)
```

translates to -> we always write new values unless the timestamp is going backwards?

mmcallister-cll: we thought about this yesterday but it could lead to more present rubberbanding. I guess if we're ok with that then it can be simplified

mxiao-cll: So the concern is that two transport returns the same timestamp with different values?

mmcallister-cll: yes

mmcallister-cll: Since this feature is currently only specific to Finage mitigation, it's very possible since their APIs send the same timestamps with mid for WS and we compute (bid+ask)/2 for REST

mmcallister-cll: That said, I expect it would likely only flip once per timestamp from WS -> REST, but it would happen for every timestamp

src/cache/response-cache/compare.ts
```diff
@@ -17,7 +17,8 @@ export class CompareResponseCache<
   // A local map to keep track of the most recent entries written to the responseCache
   // We compare with this first before comparing with value in cache
   // so that we can reduce cache reads
-  localCache: Map<string, AdapterResponse<T['Response']>>
+  localCache: Map<string, { value: AdapterResponse<T['Response']>; writtenAt: number }>
+  staleThreshold: number
```
mxiao-cll: Do you want to look into if this can be removed? The only reason this is here is to prevent overloading redis (Since we now read every time before write) 

If redis have some kind of local cache already we can kind of get rid of it.

------

Otherwise we should not over-complicate this, just set a fixed number here to expire should be fine

mmcallister-cll: there's a Redis getMany I think.. presumably using that we could fetch all with 1 request rather than 1 req per key

mxiao-cll: Was referring to https://github.com/smartcontractkit/ea-framework-js/blob/main/src/cache/redis.ts#L32

mmcallister-cll: Removed the local cache

mxiao-cll (approved):

PR #867, opened 2026-07-29
[DS-3707] Add adapter versions to EA responses

brunotm:


alejoberardino (changes requested):
The response mechanism should try to be as passthrough as possible, built on writes, so we can keep performance as well as make it easier for ourselves in the future to eventually replace the interfaces to have the EAs push instead of being polled

mxiao-cll:
Adapter version is available via the /status endpoint if you want something of the box

PR #874, opened 2026-08-13
feat(util): add validateResultNumber helper

danwilliams-cll:
Adds a reusable helper to validate numeric provider results before coercion, returning a structured 502 error for null/undefined/NaN/empty/objects/arrays/booleans/Infinity. Includes behavior-table unit tests.

### DIFF
--- a/src/util/result-number.ts
+++ b/src/util/result-number.ts
@@ -19,13 +19,6 @@ export interface ValidateResultNumberOptions {
 }
 
 /**
- * Type guard for the error branch of ValidatedResultNumber.
- */
-export const isProviderError = (
-  v: ValidatedResultNumber,
-): v is { statusCode: 502; errorMessage: string } => 'statusCode' in v
-
-/**
  * Validates that a raw value from a data provider is a usable numeric result.
  *
  * The raw value is checked BEFORE any Number() coercion so that null/undefined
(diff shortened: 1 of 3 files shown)


src/util/result-number.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  const isWrongType = typeof value !== 'number' && typeof value !== 'string'
+  const trimmed = typeof value === 'string' ? value.trim() : value
+  const num = Number(trimmed)
+
+  const isInvalid =
+    isWrongType || trimmed === '' || !Number.isFinite(num) || (!acceptZeroValue && num === 0)
+
+  if (isInvalid) {
+    logger.warn(`${errorMessage} (received: ${JSON.stringify(value)})`)
```
alejoberardino: We probably should avoid stringifying in logs, should be able to pass this out as an object (`{ msg: "asd", otherprop: obj })and it will only be parsed to a string if the log level is set to it (usually a warn will do it anyways but just as a general comment

src/util/result-number.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  acceptZeroValue?: boolean
+}
+
+/**
+ * Type guard for the error branch of ValidatedResultNumber.
+ */
+export const isProviderError = (
+  v: ValidatedResultNumber,
+): v is { statusCode: 502; errorMessage: string } => 'statusCode' in v
```
alejoberardino: This is only used for tests, should live below

alejoberardino (approved):

PR #875, opened 2026-08-13
refactor(transports): enforce resultValidator in ResponseCache base class

danwilliams-cll:
Introduces a resultValidator hook on AdapterEndpoint that runs after timestamp attachment and before the response cache write. Wires it through HttpTransport, WebSocketTransport, and SseTransport. Also propagates the validator through CompositeTransport to its child transports.

### DIFF
--- a/src/adapter/endpoint.ts
+++ b/src/adapter/endpoint.ts
@@ -109,6 +109,10 @@ export class AdapterEndpoint<T extends EndpointGenerics> implements AdapterEndpo
       inputParameters: this.inputParameters,
     })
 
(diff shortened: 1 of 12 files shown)


alejoberardino (commented):
The approach would work, my only concern (that I don't have a good answer for either) is how to enforce that this validation is always done on the transport, when users could extend them to subclasses and forget to apply the validations. Maybe we need some intermediate layer between the transport and the underlying cache that will apply them always ?

alejoberardino (approved):

PR #876, opened 2026-08-13
feat(adapter): add NumericResultEndpoint and fix LwbaEndpoint validation

danwilliams-cll:
Adds NumericResultEndpoint for single-number endpoints and moves LwbaEndpoint invariant validation from customOutputValidation (which threw a 500) to a pre-cache resultValidator that returns a 502. Depends on the resultValidator hook from #875.


src/adapter/lwba.ts
```diff
@@ -70,18 +71,53 @@ export class LwbaEndpoint<T extends LwbaEndpointGenerics> extends AdapterEndpoin
...
+
       const data = output.data as LwbaResponseDataFields['Data']
       const error = validateLwbaResponse(data.bid, data.mid, data.ask)
-
       if (error) {
         throw new AdapterLWBAError({ statusCode: 500, message: error })
       }
 
       return undefined
     }
```
alejoberardino, line 96: Only one comment: we can merge this now, but it would be nice to check if any of the adapters we have are using this, and ideally instead of keeping this validation in the read path, change those particular EAs instead

alejoberardino (approved):

PR #883, opened 2026-08-18
DF-24268 Cover bigint conversion in log censoring

alejoberardino:
Some EAs currently see [Unknown] issues from trying to log objects with bigints, and JSON serialization fails. Since logs pass through the censoring method here, we can apply handling for this case in particular in there. We default to trying to serialize normally first, since that path will drop to the native JSON serializer impl which is much faster than the JS level one.

### DIFF
--- a/src/util/logger.ts
+++ b/src/util/logger.ts
@@ -221,18 +221,18 @@ export function censor(obj: any, censorList: CensorKeyValue[], throwOnError = fa
   try {
     // JSON.stringify(obj) will fail if obj contains a circular reference or a bigint.
     stringified = JSON.stringify(obj)
-  } catch (e) {
+  } catch {
     try {
       // Retry with a bigint-safe replacer in case the failure was due to a bigint value.
       // JSON.stringify with a replacer function is slower, so we only pay that cost when
       // the fast path above actually fails, rather than on every call.
-      stringified = JSON.stringify(obj, (_key, value) =>
-        typeof value === 'bigint' ? value.toString() : value,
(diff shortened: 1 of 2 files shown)


src/util/logger.ts
```diff
@@ -219,14 +219,23 @@ export const loggingContextMiddleware = (
...
+    // JSON.stringify(obj) will fail if obj contains a circular reference or a bigint.
     stringified = JSON.stringify(obj)
   } catch (e) {
-    if (throwOnError) {
-      throw e
+    try {
+      // Retry with a bigint-safe replacer in case the failure was due to a bigint value.
+      // JSON.stringify with a replacer function is slower, so we only pay that cost when
+      // the fast path above actually fails, rather than on every call.
+      stringified = JSON.stringify(obj, (_key, value) =>
```
mxiao-cll: Why don't we just do this in the 1st pass?

alejoberardino: To avoid the performance hit unless we run into errors, since this branch is not very common doesn't make sense to slow down by default

mxiao-cll: How slow is this?

mxiao-cll (approved):

src/util/logger.ts
alejoberardino: Around twice as slow with the replacer, from a very quick poc benchmark I just ran. The call itself is fast for a small object (.1 us) but as they grow and the number of calls grow this can start becoming significant, especially on slow cpus

